PR #326, opened 2014-04-29
Autodetection for the pngcairo terminal of gnuplot.

### LINKED ISSUE #322: Feature Request: Runtime-detection for the optimal value of  $wxplot_pngcairo
On my system (Intel I7 Ultra Low voltage processor) the maxima command 
system("(gnuplot -e \'set term pngcairo\')");
needs only 50 miliseconds and returns 0 if gnuplot supports the pngcairo terminal.

Since pngcairo produces significantly higher-quality graphics than png and supports more line styles (if the "dashed" terminal is enabled, see ticket #321) adding an automatic test if we are able to use it (that automagically enables this terminal if we can use it) - would be a welcome enhancement on my side.

gunterkoenigsmann:
Closes #322.
This patch  adds an autodetection if we can use gnuplot's pngcairo terminal (produces much better graphics than the ordinary png terminal, supports the linestyles dashes and dash_dots from the current version of maxima and all linestyles the next maxima version will).

This autodetection is quite fast (50 miliseconds on my laptop). And I have updated it to fail silently (for example in the case there is still a gnuplot version out there that doesn't support pngcairo) falling back to the png terminal wxmaxima used from the beginning. 
In theory in the future we could adding a cache mechanism for the autodetection's result so it is kept for the next run of wxMaxima. But that would mean we would have to deal with cases like: "What happens if a system change exchanges a gnuplot with wxmaxima by a version without" or ""What happens if an user uses the same home directory on two computers only one of which supports pngcairo." Not making any assumptions about the system should IMHO be the better option.

---
Compatibility: Other than the increased graphics quality an users that don't try to use other linestyles than "solid" or "dots" (the png terminal ignored all other line styles) won't see any difference when my patches are applied.

### DIFF
--- a/data/wxmathml.lisp
+++ b/data/wxmathml.lisp
@@ -1242,7 +1242,7 @@
   (cond
    ((like (simplify
 	   (mfunction-call $system
-			   '"(gnuplot -e 'set term pngcairo')"))
+			   '"(gnuplot -e 'set term pngcairo dashed')"))
 	  0.)
     (setq $wxplot_pngcairo t)))
   )
--- a/data/wxmathml.lisp
+++ b/data/wxmathml.lisp
@@ -1238,7 +1238,7 @@
 (defmvar $wxplot_pngcairo nil)
 
 (catch
-   (setq $wxplot_pngcairo t)
+   (setq $wxplot_pngcairo nil)
   (cond
    ((like (simplify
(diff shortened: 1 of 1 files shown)


gunterkoenigsmann:
Did add such a way in commit 7d61b3b72271673261f0a94f1439eb1985a3483c.

PR #417, opened 2015-02-11
Update Japanese translations

matcha-zaq:


matcha-zaq:
This is a Japanese translations update.

PR #449, opened 2015-04-07
Update Ukrainian translation

yurchor:
The update is tested with msgfmt -vc and on the running application (in vivo).

Many thanks for merging this request.


gunterkoenigsmann:
Thanks to you!

    Gunter.

PR #744, opened 2016-07-10
Create kab.po

belkacem77:
Adding a localization file for Kabyle language

### DIFF
--- a/locales/kab.po
+++ b/locales/kab.po
@@ -1,0 +1,5771 @@
+# This file is distributed under the same license as the PACKAGE package.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER.
+# Eric <ericdelevaux@gmail.com>, 2004.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: fr\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2016-05-07 08:48+0200\n"
+"PO-Revision-Date: 2016-07-10 20:53+0100\n"
+"Last-Translator: Belkacem Mohammed <belkacem77@gmail.com>\n"
+"Language-Team: kabyle <kab@li.org>\n"
+"Language: kab\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Poedit 1.8.8\n"
+
+#: ../src/wxMaxima.cpp:4348
+#, c-format
+msgid ""
+"\n"
+"\n"
+"wxWidgets: %d.%d.%d\n"
+"Unicode support: %s"
+msgstr ""
+"\n"
+"\n"
+"wxWidgets: %d.%d.%d\n"
+"Support d'Unicode: %s"
+
+#: ../src/wxMaxima.cpp:4361
+msgid ""
+"\n"
+"Lisp: "
+msgstr ""
+"\n"
+"Lisp : "
+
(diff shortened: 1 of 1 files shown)


belkacem77:
We are going to translate wxmaxima into Kabyle language. Lot of strings are translated. The project is going on.

gunterkoenigsmann:
Then welcome on board! Support is always appreciated and it seems you did a great job. The statistics msgfmt outputs are:

786 translated messages, 83 fuzzy translations, 199 untranslated messages.

If I didn't make a mistake right now the next nightly build should already contain your language file and the option to select you language.

Kind regards,

   Gunter.

gunterkoenigsmann:
Dear Andrej,

> We are going to translate wxmaxima into Kabyle language. Lot of strings
> are translated. The project is going on.

I have tried to add this language. But it seems that the travis build
for Linux (but not for Mac computers) fails with the following message:

ConfigDialogue.cpp:66:3: error: ‘wxLANGUAGE_KABYLE’ was not declared in
this scope

I have put a

#ifdef wxLANGUAGE_KABYLE

around the language definition. But as wxLANGUAGE_KABYLE is part of an
enum I think this ifdef always returns false.

Do you have an idea what to do?

Thanks a lot,
and kind regards,

    Gunter.

andrejv:
My best idea is to use wxCHECK_VERSION if you know when the support for wxLANGUAGE_KABYLE was added.

andrejv:
I think you need at least version 3.0.1

belkacem77:
Hi,
Thanks for your help/support. I will announce it for my community. A Facebook page is created to discuss Kabyle Localization.

Is there any online translation site to manage Maxima localizations? Such as Transifex, Launchpad..

Ragards;
 M.Belkacem

gunterkoenigsmann:
On 11.07.2016 11:32, Belkacem Mohammed wrote:
> Hi,
> Thanks for your help/support. I will announce it for my community. A
> Facebook page is created to discuss Kabyle Localization.
> 

Cool! The community always is important.

> Is there any online translation site to manage Maxima localizations?
> Such as Transifex, Launchpad..

Unfortunately I haven't been able to find any such site that really
supports SourceForge. The devs seem not to like the idea of a
translation site that involves much manual work too much.  ...And for
the >1000 pages manual I never found a good way of maintaining a
translation at all. I thought that I once had seen a program that
converts every paragraph to a "string" for gettext so gettext will show
the translators which paragraphs have changed and need to be revised.
But I never found such a thing when I searched for it again ;-(

But sending in patches works so there is a way to add localization
there, too.

---

Currently wxMaxima detects several of maxima's error messages by
actually comparing a message from maxima with a few candidates for
potential error messages. If maxima's error messages are translated we
need to add the ability to recognize the translated error messages to
wxMaxima, too.

---

If the nightly build infrastructure uses a new enough wxWidgets the next
nightly build should be configurable to use the new language.

Thanks again,
and
Kind regards,

  Gunter.

gunterkoenigsmann:
Thanks a lot!

In the meantime I have merged some performance improvements into the master.

Kind regards,

 Gunter.

gunterkoenigsmann:
Dear Mohammed,

wxMaxima seems to complain at startup if the language of the last
nightly window build has been set to "Kabyle": The string "%.3f" seems
to use a different decimal separator than the locale. I assume that If
in Kabyle it is more common to use a comma in floating-point numbers
replacing the string by "%,3f" fixes the problem. If not we have to
search at a different place.

---

For asking questions about localizing maxima itself the maxima-discuss
list at maxima.sourceforge.net should be the best place. Perhaps they
have new ideas about localization services.

Kind regards,

 Gunter.

PR #747, opened 2016-07-15
Fix filename and .desktop reference for appdata.xml.

badshah400:
* Change appdata filename to be consistent with .desktop file name
* The reference to the .desktop file in appdata.xml was to
  wxmaxima.desktop, while the installed .desktop file was
  called wxMaxima.desktop, fix this.

The renaming of the appdata file is not strictly required but recommended for distro-specific package managers to pick up the appdata file consistently [see http://dominique.leuenberger.net/blog/2016/07/how-to-fix-my-application-does-not-show-up-in-software-centers/]. Comments welcome.

Thanks!

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -7,7 +7,7 @@ bashcompletion_DATA = wxmaxima
 EXTRA_DIST = wxmaxima tips.txt wxmathml.lisp wxmaxima.png wxmaxima.svg Info.plist.in PkgInfo autocomplete.txt wxmaxima.manifest
 
 appdatadir = $(datarootdir)/appdata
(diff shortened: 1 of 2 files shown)


gunterkoenigsmann:
Why did this work? Hmmm.... and I suppose we need to rename data/wxmaxima.xml in order to make this work.

gunterkoenigsmann:
Ok... ...the rename was there but I somehow failed to see it. Thanks a lot for this patch!

badshah400:
Thanks for accepting my patch so quickly.

gunterkoenigsmann:
You are welcome!

PR #786, opened 2016-10-16
Enable better translation of "save changes" dialog

minoki:
Use formatting function to include the file name in the message, instead of string concatenation.


gunterkoenigsmann:
Great idea! Thanks!

PR #790, opened 2016-10-18
Update Japanese translation

minoki:
I've updated Japanese translation of wxMaxima.

Before: 776 translated messages, 77 fuzzy translations, 255 untranslated messages.

After: 903 translated messages, 205 untranslated messages.


gunterkoenigsmann:
Thanks a lot!
and kind regards,
    Gunter.

PR #865, opened 2017-03-12
Fix a typo that is an algebra bug.

KubaO:
Clang warns about these things :) PVS Studio does too.


gunterkoenigsmann:
That solves a complete mystery! Thanks a lot!

PR #875, opened 2017-04-10
Updated Italian translation unstranslating untranslatable accel key names

ciampix:
Accel key names seems to be untranslatable since when I start wxMaxima it complains with a series of:

13:31:39: Debug: Unknown accel modifier: 'maiusc'

Probably we should open a bug report for making translatable the accel key names...


gunterkoenigsmann:
Don't bc know what wxWidgets did there: had the same problem with some german accelerators, too.

PR #900, opened 2017-07-06
Update tr.po

TufanSirin:
I am tired. Tomorrow I will complete. :) I wish. good night


gunterkoenigsmann:
Thanks a lot! I've corrected a few places where your text editor seems to have a different idea about how to escape " signs than msgmerge.

PR #915, opened 2017-08-27
Better TeX

astyonax:
So, 
as I said in the [issue message](https://github.com/andrejv/wxmaxima/issues/912), here is the pull request to review the code changes.

At fist I was confused on the way the regular expressions work in the wxWidgets, but then I realized that 
* The function `MathCell::ListToTex()` does not see all the output text when the user clicks on `Copy to LaTeX`.  You can convince yourself with a call to wxLogMessage(retval) in the same function: this will open the log window with the content of `retval`.
* For this reason, I moved all the regex code that was in there to a new function `BetterTeX` which lives in the corresponding files (.cpp & .h).
* The function is now called from the function `MathCtrl::CopyTeX()`.

At the moment not all the necessary regex are implemented, but you can already see their work in the file `test/regex.wxmx`


astyonax:
I agree that build.sh should not be in the master trunk. It has to be removed when merging the BetterTeX branch to master.
Also the white spaces changes: that's due to the editor, and I'll remove those changes in one of the next commits. 

I'll finalize these details in the next commits.

astyonax:
Hi, 
I think I have a reasonably working version. 
I tested the output in the file test/regex.wxmx .

I did not just implement the regular expressions,
but I also tweaked the ToTeX functions of many type of cells to 
avoid the insertions of redundant braces.

What do you think on the current status?

ciao

gunterkoenigsmann:
This is a big improvement to TeX export.

PR #953, opened 2018-01-19
Modified scaling of toolbar/statusbar icons.

LennartF22:
First of all, I rewrote the code to use the internal scaling mechanism of the icons provided by the wxArtProvider (if available).

Then I tried to load the .svg files instead of the .png files to realize, that wxWidgets can't really handle external .svg files without additional libraries. Therefore I simply increased the dimensions of the icons to 192x192 px, which didn't increase the file size of the images by a huge amount.

The quality of the scaled icons provided by the wxArtProvider was increased by a lot on my linux system. The quality of the icons when loaded from the .png files also increased to some extend. Obviously this is subjective ;)

After all the optimal solution would be to generate bitmaps from .svg files as they are loaded, but I don't know whether the effort (additional libraries) would be worth it...


gunterkoenigsmann:
Great work! Thanks a lot!

PR #958, opened 2018-01-21
Sub Sup Cell MathML

LennartF22:
I don't know if MathML is (still) used for anything, but while having a look at the sub sup cell issue, I noticed that index and exponent are the wrong way around for MathML aswell.


gunterkoenigsmann:
There is a very active community that looks if any application supports MathML. Also MS Word supports it. But Libreoffice (even if it has a MathML import) doesn't recognize MathML on Drag-and-drop. Which is why I had to implement RTF - which (as I don't use Word and the RTF standard is hard to read) won't be perfect.

The MathML community also wants every application to support MathML on Drag-and-drop. I am sceptical if that is actually possible in most cases: For example how do we decide if abs() or cabs() is needed? But as the MathML we produce always contains the maxima code the MathML code corresponds to as

     <annotation encoding="application/x-maxima">

field we one day could at least start to interpret our own MathML on drag-and-drop in case that someone used it

PR #960, opened 2018-01-22
Changed the Maxima path escaping stuff to work properly.

LennartF22:
Instead of escaping spaces, I rewrote the code to escape quotation marks and to enclose the path itself by quotation marks. Now everything seems to work properly on Windows and Linux.


gunterkoenigsmann:
Thanks a lot! If you want to from now on you can commit corrections directly to the wxMaxima repo.

PR #961, opened 2018-01-22
Fixed print margins (most of it).

LennartF22:
The new margins for printing had to be calculated after setting the new zoom factor. A problem that's still occurring is that e.g. titles and sections have some additional indent, which the StyleText() method isn't taking care of. I fixed this by adding some additional margin when calculating the new client width, even though this isn't optimal, obviously. I also tried out a different approach before ([PrintMargins](https://github.com/LennartF22/wxmaxima/tree/PrintMargins)), and the problem described above didn't occur, but that wasn't necessarily a "tidy" fix either.

Unfortunately, I don't think that I have the time to do a tidy fix, as I think that such a fix would require some serious work, but I don't know.

This is how things look right now (independent of the dpi and identical on Linux and Windows, with exception of the font): [softwrap.pdf](https://github.com/andrejv/wxmaxima/files/1653567/softwrap_linux.pdf)

### DIFF
--- a/src/EditorCell.cpp
+++ b/src/EditorCell.cpp
@@ -1669,7 +1669,7 @@ bool EditorCell::HandleSpecialKey(wxKeyEvent &event)
         if ((!cursorJump) || ((cursorAtStartOfLine) && (!autoIndent)))
           m_positionOfCaret = BeginningOfLine(m_positionOfCaret);
       }
-      //StyleText();
+      StyleText();
       break;
 
     case WXK_DELETE:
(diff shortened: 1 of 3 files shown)


gunterkoenigsmann:
In the meantime Andrej has given you write access to the main wxMaxima repo => If you want you can now commit directly there, instead.

LennartF22:
Looks like I don't have write access yet, because I wasn't able to create a new branch this afternoon. Also, when I try to edit a file through the browser, it says the following:
> You’re editing a file in a project you don’t have write access to. Submitting a change to this file will write it to a new branch in your fork LennartF22/wxmaxima, so you can send a pull request.

andrejv:
@LennartF22 you should have received and invitation email from github inviting you to join the project. You need to accept the invitation to receive the permissions (https://github.com/andrejv/wxmaxima/invitations).

LennartF22:
Oh, I must have overlooked that email, as I received a dozen emails from GitHub that day 😄

PR #1049, opened 2018-11-07
fix wxgtk2 and wxgtk3

mar04:
Based on changes suggested by @kosilin
However I'm not sure if antialiassing works properly in wxgtk3 now: https://imgur.com/5J66j1D
on the left wxMaxima 18.02 on the right current master with this patch. One can see difference when looking at parenthesis.

### DIFF
--- a/src/MathCtrl.cpp
+++ b/src/MathCtrl.cpp
@@ -386,7 +386,7 @@ void MathCtrl::OnPaint(wxPaintEvent &WXUNUSED(event))
 
   wxGCDC antiAliassingDC(dcm);
 
-  #ifdef __WXGTK__
+  #if defined(__WXGTK__) && !defined(__WXGTK3__)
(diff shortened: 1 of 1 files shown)


gunterkoenigsmann:
Imgur has jpeg-compressed the images so the difference isn't visible any more. But as we still use a wxGCDC the only case we won't have antialiasing now is if wxWidgets or Gtk is broken => the antialiasing problem will have to be resolved on their side, not on ours.

kosilin:
Thank a lot once more! Antialiasing at my environment seems to be present: https://www.dropbox.com/s/iecvz5qafhwua5d/antialiasing.png?dl=0 , but it doesn't respect system settings (rgb). But it issue isn't related only to wxMaxima or wxGTK3 it's related rather to particular way the window content is rendered.

PR #1081, opened 2019-01-24
Fix AppData file: Make only one default screenshot

scx:
There should be only one default screenshot.

Property `type="default"` is intended to indicate the application's primary screenshot.

This patch has already been applied to the flatpak package.
https://github.com/flathub/flathub/pull/801#issuecomment-457083756


gunterkoenigsmann:
Thanks a lot!
If you prefer to do so if you checkout the git repo using the ssh method you can directly commit to the repo - at least if you have told github about your public ssh key.

PR #1105, opened 2019-04-06
add copy to matlab

Murmele:
I'm calculating my matrices in wxmaxima and use them in matlab for simulation. Everytime I copy the matrices, I have to do some changes, because the copied matrix has not the same format like in matlab. For this it's nice to copy directly matlab code and just paste it in matlab.
(Maybe an alignment can be made, then it looks like in matlab symbolic toolbox, but the matrices are then much bigger)

### DIFF
--- a/src/AbsCell.cpp
+++ b/src/AbsCell.cpp
@@ -160,6 +160,15 @@ wxString AbsCell::ToString()
   return s;
 }
 
+wxString AbsCell::ToMatlab()
+{
+  if (m_isBrokenIntoLines)
+	return wxEmptyString;
+  wxString s;
+  s = wxT("abs(") + m_innerCell->ListToMatlab() + wxT(")");
+  return s;
+}
+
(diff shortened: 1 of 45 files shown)


gunterkoenigsmann:
Thanks a lot!

gunterkoenigsmann:
Thanks a lot!

Murmele:
Thank you for merging so fast!

gunterkoenigsmann:
I am no octave expert but anything that makes Maxima interface better with the numeric calculation program is highly welcome.

Murmele:
Hmm at the moment I don't have an idea, but if you like I can ask in the octave mailing list if there are interests.

gunterkoenigsmann:
Octave is free, but seems to understand all the commands your "copy to matlab" feature generates.

Murmele:
ah you mean like this. Yes I think there must be changed anything. Maybe we can name it "copy to matlab/octave"?

PR #1126, opened 2019-05-18
Fix AppData file

scx:
AppData validation has become more strict:
 * longest allowed caption is 50 chars
 * `<li>` cannot contain a hyperlink


gunterkoenigsmann:
...and feel free to add yourself to the list of authors in AUTHORS and SRC/wxMaxima.cpp: I've always forgotten to maintain these lists.

PR #1127, opened 2019-05-18
Enhance AppData file

scx:
Enhance AppData file:
 * Add `provides` tag (binary)
 * Add `translation` tag (gettext)

See also:
https://github.com/flathub/flathub/wiki/AppData-Guidelines#translations
> If the app uses gettext `.mo` or Qt `.qm` files, you’ll need to provide the prefix of these files with a `<translation/>` tag

### DIFF
--- a/data/io.github.wxmaxima_developers.wxMaxima.appdata.xml
+++ b/data/io.github.wxmaxima_developers.wxMaxima.appdata.xml
@@ -49,6 +49,10 @@
     <kudo>UserDocs</kudo>
   </kudos>
   <launchable type="desktop-id">io.github.wxmaxima_developers.wxMaxima.desktop</launchable>
+  <provides>
+    <binary>wxMaxima</binary>
+  </provides>
+  <translation type="gettext">wxMaxima</translation>
   <content_rating type="oars-1.0">
     <content_attribute id="social-info">mild</content_attribute>
   </content_rating>


gunterkoenigsmann:
Does the AppData file need to be translated? Or is there a good way to do so? I could provide german.

scx:
@gunterkoenigsmann 

> Does the AppData file need to be translated? Or is there a good way to do so? I could provide german.

As you can see, the `<translation>` tag refers to `share/locale/<lang>/<prefix>.mo` files. However, you can apply translation for the AppData file itself. It is not strictly required, but you can get "bonus points" for this.
First of all, the `appstream-glib` tool provides 3 levels of validation: `validate-relax`, `validate` and `validate-strict`. Only the last one requires translations for `<name>`, `<summary>` and `<description>` tags.
Flathub requires AppData to pass `validate-relax` validation... or so it was until this week. It looks like they now require the `validate` level.
https://github.com/flathub/flathub/wiki/App-Requirements#Appstream
https://flathub.org/builds/#/builders/1/builds/290

Anyway, if you want, you can provide translations for mentioned tags directly in the AppData file.
https://github.com/Beep6581/RawTherapee/blob/01894609ce10453e420e78fa075ef10399e7f0c2/com.rawtherapee.RawTherapee.appdata.xml#L6-L8

Or you can use `gettext` to do this for you.
https://gitlab.gnome.org/scx/ocrfeeder/commit/dd57b0e0f646d328452f9fa43104e7163ba9abed
https://gitlab.gnome.org/scx/ocrfeeder/blob/dd57b0e0f646d328452f9fa43104e7163ba9abed/resources/ocrfeeder.appdata.xml.in#L11-13

You can also leave it as it is. Your choice.

PR #1135, opened 2019-06-14
Fix incorrect use of ^ operator

jwakely:
`2^32` equals `34`, which does not do what this code wants.


gunterkoenigsmann:
I *knew* I had made this wrong in one place. Thans a lot!

PR #1146, opened 2019-08-08
Fix AppData file

scx:
Fix AppData file:
 * Fix indents
 * `<li>` cannot end in `.`


scx:
The build failed, but it wasn't related to this patch.

PR #1149, opened 2019-08-18
Handle partial writes and errors in SendMaxima

stevenjm:
Without this patch, partial writes go unnoticed by wxMaxima, resulting in Maxima seeing a partial command followed immediately by whatever the next command to be written happens to be. On OpenBSD, this seems to happen more often than on other OSes, as I consistently see this message on startup:

```
Maxima encountered a Lisp error:
 Unexpected end of file on #<string-input stream from "(:lis...">.
Automatically continuing.
To enable the Lisp debugger set *debugger-hook* to nil.
Maxima encountered a Lisp error:
 The function READ-WXMAXIMA-VERSION is undefined.
Automatically continuing.
To enable the Lisp debugger set *debugger-hook* to nil.
```

This is caused by wxMaxima's startup commands being incompletely written.

Regardless of which platforms this bug manifests on, this is the documented behaviour of the wxSocketBase::Write call:

https://docs.wxwidgets.org/3.1/classwx_socket_base.html#a7676f1821d2e17db95de2438346355cd
> Use LastWriteCount() to verify the number of bytes actually written.

I am uncertain whether exit(1) is the correct way to handle errors in this code, but I can't think of a better approach. This method is called very early after starting Maxima (from SetupVariables), so trying to automatically recover by restarting Maxima could easily lead to an infinite loop of Maxima restarts. I think it is better to fail hard than to risk such a loop.

### DIFF
--- a/src/wxMaxima.cpp
+++ b/src/wxMaxima.cpp
@@ -746,7 +746,25 @@ void wxMaxima::SendMaxima(wxString s, bool addToHistory)
       else
         StatusMaximaBusy(waiting);
       wxScopedCharBuffer const data_raw = s.utf8_str();
-      m_client->Write(data_raw.data(), data_raw.length());
+      const void *buf = data_raw.data();
+      wxUint32 len = data_raw.length();
+      wxUint32 last;
+      while (1) {
+        if (!m_client->WaitForWrite()) {
+          std::cerr << "Timeout waiting for Maxima socket to be writable" << std::endl;
+          exit(1);
+        }
+        m_client->Write(buf, len);        
+        if (m_client->Error()) {
+          std::cerr << "Error writing to Maxima" << std::endl;
+          exit(1);
+        }
+        last = m_client->LastWriteCount();
+        if (last == len)
+          break;
+        buf = (const void*)((uintptr_t)buf + last);
+        len -= last;
+      }
       m_statusBar->NetworkStatus(StatusBar::transmit);
     }
   }


gunterkoenigsmann:
I never thought partial writes were actually possible, as long as maxima doesn't crash, that is. ...and never have seen one...

gunterkoenigsmann:
For maxima we cannot do anything if it fails to read data. But I've modified the code to keep the worksheet open so the user can save the file, even if maxima has failed.

stevenjm:
> For maxima we cannot do anything if it fails to read data. But I've modified the code to keep the worksheet open so the user can save the file, even if maxima has failed.

Thanks! I've tested on OpenBSD with the code from master and all works fine.

gunterkoenigsmann:
One last question, though: Do you know a reliable way to provoke a timeout so I can test future modifications of the code?

stevenjm:
The only way I can find to provoke a timeout is by blocking return traffic on the TCP socket to Maxima **before** starting wxMaxima, because I can only reproduce partial writes with the commands wxMaxima sends on startup and not once it is already running.

Adding this block once wxMaxima is already running allows the write to succeed (into the socket buffer), but wxMaxima then hangs forever awaiting a reply from Maxima, instead of triggering the write timeout. This may be a peculiarity of OpenBSD's socket implementation; perhaps the socket buffer is enlarged dynamically once traffic begins to pass.

Whatever the reason, putting this in /etc/pf.conf, reloading the firewall, and then starting wxMaxima works for me:

```
block drop in quick on lo proto tcp from any port { 40100 40101 40102 40103 } flags /SA no state
pass in quick on lo no state
pass out quick on lo no state
```

This allows the TCP handshake between Maxima and wxMaxima to proceed as normal, and then applies a one-way block on traffic originating from wxMaxima. After the first partial write is encountered, WaitForWrite then blocks waiting to send more data, which of course never happens, and I get a timeout.

Sadly this isn't a very automatable or portable test, but every other approach I've tried either triggers a write error instead of a timeout, or else loops forever reading a response instead.

stevenjm:
So I don't think this change was entirely correct, but I think a correct fix requires a little more thought. The implementation of `WaitForWrite` seems to wait for an **event** indicating the socket has become ready for writes, which means that calling it on a socket that is already ready for writes might time out even if the socket is writable, which is the behaviour I'm now seeing. (The documentation does say that it shouldn't be called in such cases, but doesn't explain why, and this makes it impossible to use here without race conditions.)

For now I have things working locally by just sleeping for 10 ms each time a `wxSOCKET_WOULDBLOCK` error is encountered. In principle, we should be able to just call `WaitForWrite` with a small timeout and try again regardless of success, but when I tried that it ignored the timeout I specified.

wxWidgets just passes that timeout along to the underlying implementation (GTK+, in my case). Properly understanding this problem requires some study of whether GTK+ is doing the right thing, and perhaps whether this happens with other wxWidgets implementations. I'm willing to invest some study into this, but probably won't have much time until next weekend.

In the meantime, I'm not sure if we should revert this change -- it definitely causes hangs on my OpenBSD system, but I'm not sure if that will also be the case elsewhere, since there weren't problems with partial writes elsewhere to begin with.

There is a related race condition here where `SendMaxima` can be called again from another thread while it is already waiting to write in a different thread. There should probably be a mutex around some of this code. I'll look into that as well; ideally, the way writes are done can be refactored to resolve both issues.

gunterkoenigsmann:
WxMaxima only uses a single thread. But you are right: We should append the data we want to send to a wxMemoryBuffer instead of replacing something. Hope I corrected it the right way now.

gunterkoenigsmann:
Ok... ...now we do the following:

- All data we want to send is appended to a wxMemoryBuffer
- If the wxMemoryBuffer was empty before that we send its contents to maxima
- when we receive the signal that we can send new data to maxima we ask how much data we sent and (if the buffer is longer than the total amount of data sent) initiate another send.
 - and if the connection is lost we kill maxima and leave it to the respawn mechanism to restart it or to give up.

What I don't know is if any of this handles timeouts. I guess they are either handled as a "connection lost" or as a "we have sent 0 bytes and are ready to send bytes again". Both cases would (hopefully) be handled correctly. But I have not found any documentation telling if I am right.

stevenjm:
Thanks! This at least seems to work in the normal case on my machine.

I'll hopefully find some time to test this with timeouts soon.

stevenjm:
I can't seem to trigger a timeout condition with this -- whatever I try either doesn't stop communication with Maxima or causes it to hang indefinitely (or for as long as I had the patience to wait). But I don't think it matters very much, timeouts aren't a very likely failure mode when talking to Maxima over a local socket.

Thanks again for this fix, and sorry for the trouble!

gunterkoenigsmann:
If it fixes bugs it is always worth the trouble. And we have fixed 2 of 3 things:

- We now handle partial writes and
- we have increased wxMaxima's speed.

If we ever find a way to handle timeouts, as well, that would be great. But as you have stated this currently doesn't seem to be a big problem.

gunterkoenigsmann:
On the other hand we might have already resolved the whole problem, in a way: If I understand it right if the communication with maxima blocks currently wxMaxima's gui doesn't block and wxMaxima retries indefinitely to communicate saving us from setting a timeout - that always would always have the wrong duration, anyway.

Issue #1031, opened 2018-09-20
Bug with blank spaces


SengerM:
Hello,
I have made a question in the discussion forum, but making some tests I think I have found a bug. I am sorry if it is not. I am having a very strange behavior with blank spaces. The image in [this link](https://image.ibb.co/fwkuWe/imagen.png) illustrates the problem I am having. Before the definition of 'C' it seems that the "number 0" is being interpreted as "the variable 0". This happens for other numbers as well, as can be seen in [this link](https://image.ibb.co/inkYPz/imagen.png). I had also problems with symbols having the same name but being considered as different by wxMaxima.

I am running wxMaxima 18.02.0 ([this version](https://image.ibb.co/jh6RSK/imagen.png)) in Xubuntu 18.04.

If I execute a script in maxima from terminal containing this:

	A: 1=1;
	B: 2=2;
	C:3=3;
	D: 4=4;

	A^2;
	B^2;
	C^2;
	D^2;

everything works as expected.

gunterkoenigsmann:
Does the issue disappear if you only press the space bar without pressing shift? 

Shift+Space is interpreted as a non-breakable space. Not that I have introduced a big here...

SengerM:
I didn't knew about Shift+Space. I have introduced all the spaces just as ordinary ones. I have made a test using Shift+Space and [this](https://image.ibb.co/bPtAMe/imagen.png) is what I get.

gunterkoenigsmann:
As I cannot reproduce the bug on my computer: Does it still appear if you use the maxima + wxMaxima version from here: https://code.launchpad.net/~maxima-developers/+archive/ubuntu/maxima-nightly

It might be necessary to uninstall maxima+wxMaxima before installing the nightly build.

My guess is that the bug might already have been fixed, perhaps accidentally as I don't remember ever having read about it.

Thanks a lot,
Gunter.

leon-vv:
I can confirm that this bug is currently in Debian Testing but is not present anymore in the nightly build.

SengerM:
I have installed that version and it seems the problem is gone. I'll keep on testing and if something happens I'll let you know.

Thanks!

gunterkoenigsmann:
We urgently need a new Maxima+wxMaxima release.

jlindgren90:
[...]

PR #1157, opened 2019-09-30
In EditorCell::HandleOrdinaryKey, do not convert Shift+Space to nonbreaking space

robert-dodier:
This PR contains 1 commit which reverts the behavior which was introduced in commit 20e7e19. It appears that users encountered problems with Shift+Space --> nbsp, as discussed in issue #1031 (https://github.com/wxMaxima-developers/wxmaxima/issues/1031).

I believe that EditorCell::HandleOrdinaryKey is the only place in the code in which Shift+Space is converted to nbsp.

This PR only changes the behavior of EditorCell::HandleOrdinaryKey, it does not change any other code related to nonbreaking spaces (e.g. to convert nonbreaking space to tilde for TeX output) and it does not change any of the locales strings which mention Shift+Space --> nbsp, and it does not change the Tip Of The Day which mentions it.

### DIFF
--- a/src/EditorCell.cpp
+++ b/src/EditorCell.cpp
@@ -2326,9 +2326,6 @@ bool EditorCell::HandleOrdinaryKey(wxKeyEvent &event)
 
     chr = event.GetUnicodeKey();
 
-    if (event.ShiftDown())
-      chr.Replace(wxT(" "), wxT("\xa0"));
-
     m_text = m_text.SubString(0, m_positionOfCaret - 1) +
              chr +
              m_text.SubString(m_positionOfCaret, m_text.Length());


gunterkoenigsmann:
Thanks a lot for this pull request!

Seeing where the problem lies I've fixed it on the other side using 8e9746d2b49f1b994b5050e0db029bac5c4e21ba: I still allow the user to use non-breaking spaces, but convert them to ordinary spaces before sending them to maxima. I'm still not completely sure if I should exempt non-breaking spaces within strings from this search-and-replace command: If I replace all spaces unconditionally sooner or later there will be a bug report that it is hard to make maxima generate files with non-breaking spaces: Maxima's userbase is big enough for this to be guaranteed to happen. But if we leave non-breaking spaces in strings there will be bug reports that strings cannot be interpreted correctly. I personally believe the latter is bound to happen more times. But I may be wrong...

rknop:
If there are nbsp's anywhere, *please* make them look different on the screen.  Shift-space being nbsp right now is a big problem because 34 years of typing on computers (for me, and I'm sure for others) have trained us that if your thumb lingers on the shift key after a : or a ), it doesn't matter; as a result, careful typing slows us *way* down, to the point that I consider the current release of wxmaxima unusable.  But, the bigger problem is that there's no way to tell, and sometimes the error message (or the behavior, as sometimes there aren't any) doesn't make clear what your real issue is.

If you're going to have non-breaking spaces anywhere, then on the input side of things, they need to look different from spaces somehow, so that folks looking at them have a hope of seeing what's there.

jlindgren90:
I would also much prefer that Shift-Space does not add non-breaking spaces in the first place, rather than silently adding them and then silently removing them to "fix" the maxima input later.

PR #1190, opened 2019-12-24
Update Russian translation

OlesyaGerasimenko:


gunterkoenigsmann:
Thanks a lot! And - in case that you celebrate christmas on the same day I do - happy Christmas!

PR #1212, opened 2020-02-23
Fix spelling of Xcode -- only 'X' should be capitalized.

lanurmi:


gunterkoenigsmann:
Thanks a lot! Not being a mac user I guess I will have made many more mistakes for this platform, though...

PR #1228, opened 2020-03-20
Update Ukrainian translation

yurchor:
Preliminary update, incomplete, will be finished asap. Tested for validity with msgfmt -vc.

Thanks in advance for merging.


locales/wxMaxima/uk.po
```diff
@@ -2867,15 +2896,15 @@ msgstr "Текст прикладу"
 
 #: /home/gunter/src/wxmaxima/src/ConfigDialogue.cpp:956
 msgid "Examples:"
-msgstr ""
+msgstr "Приклад:"
```
Kvaz1r: Приклади?

locales/manual/uk.po
```diff
@@ -60,52 +62,52 @@ msgstr ""
...
 #: /home/gunter/src/wxmaxima/info/wxmaxima.md:23
 msgid "[Troubleshooting](#Troubleshooting \"Troubleshooting\"):"
-msgstr ""
+msgstr "[Усування вад](#Troubleshooting \"Усування вад\"):"
 
 #. type: Plain text
 #: /home/gunter/src/wxmaxima/info/wxmaxima.md:27
 msgid "What to do if wxMaxima does not work as expected"
-msgstr ""
+msgstr "Що робити, якщо wxMaxima працює неправильно"
```
Kvaz1r: Maybe better 'неочікувано' instead of 'неправильно'?

locales/wxMaxima/uk.po
yurchor: Sure. Thanks for the spotting.

locales/manual/uk.po
yurchor: I do not think so. It depends on the wrong and not unexpected behavior. We cannot fulfill all the expectations.

Kvaz1r: Yes, but usual user also can be wrong about does the actual behaviour is really wrong or just unexpected. Like any usual issue template(i.e. actual(фактична) behaviour and expected(очікувана) behavior). Sorry for bothering if I am missed context.

yurchor: It is a description of the "Troubleshooting" chapter.

http://arcoramexico.net/wxmaxima-16.12.2/info/wxmaxima.html#Troubleshooting

This chapter describes purely erroneous behavior which has nothing to do with "unexpected" things. It does not even contain links for bug reporting or something really indeterministic. Just a list of common bugs.

gunterkoenigsmann:
This is the first time I used the web interface for merging. Hope I didn't break anything. If I did please protest.

gunterkoenigsmann:
Ok... ...it seems like I broke the translation. But it should now be unbroken again.

PR #1233, opened 2020-03-23
Enhance mechanism to find gnuplot

MSoegtropIMC:
This PR enhances the mechanism to find gnuplot:

- Centralised two different places in the code searching for gnuplot into Directorystructure.cpp
- Added a Macports specific search path for OSX
- Added a mechanism to easily patch the MacPorts root folder to the actual prefix chosen on each system
- Always set m_gnuplotcommand to a path of gnuplot in case gnuplot can be found and do this only once, in the constructor of wxMaxima and when settings are read in.

### DIFF
--- a/src/Dirstructure.cpp
+++ b/src/Dirstructure.cpp
@@ -29,11 +29,14 @@
   DirStructure informs about what directory is to be found where on the current system.
 */
 
-
 #include "Dirstructure.h"
 #include <wx/filename.h>
 #include <wx/dir.h>
 #include "Version.h"
+
+// This macro is used to mark the macports prefix, so that it can be easily patched
+// The MacPorts port file for wxMaxima uses this, so don't remove this!
+#define OSX_MACPORTS_PREFIX "/opt/local"
 
 Dirstructure::Dirstructure()
 {
@@ -214,7 +217,7 @@ wxString Dirstructure::MaximaDefaultLocation()
     return maximaLocation;
 
   wxLogMessage(wxString::Format(notFound,maximaLocation.utf8_str()));
-  maximaLocation = "/opt/local/bin/maxima";
+  maximaLocation = OSX_MACPORTS_PREFIX "/bin/maxima";
   if (wxFileExists(maximaLocation))
     return maximaLocation;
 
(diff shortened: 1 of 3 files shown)


src/wxMaxima.cpp
```diff
@@ -2640,8 +2640,8 @@ void wxMaxima::ReadVariables(wxString &data)
             }
             if(name == "gnuplot_command")
             {
-              m_gnuplotcommand = value;
-              wxLogMessage(wxString::Format(_("Gnuplot can be found at %s"),value.utf8_str()));
```
gunterkoenigsmann: This is the only place we can improve the thing in: Your heuristics for finding gnuplot is fine. But we shouldn't overwrite m_gnuplotCommand if we can execute the gnuplot maxima told us to use: The user might have two versions of gnuplot installed and might have customized maxima to use the only one that works.

The following piece of code from wxMaxima.cpp should do half of the task of verifying if the gnuplot maxima offers is somewhere in the search path:

    // Find gnuplot
    wxString gnuplot_binary;
    if(wxFileName(m_gnuplotcommand).IsAbsolute())
      gnuplot_binary = m_gnuplotcommand;
    else
    {
      wxPathList pathlist;
      pathlist.Add(wxFileName(wxStandardPaths::Get().GetExecutablePath()).GetPath());
      pathlist.Add(wxFileName(wxStandardPaths::Get().GetExecutablePath()).GetPath()+"/../");
      pathlist.Add(wxFileName(wxStandardPaths::Get().GetExecutablePath()).GetPath()+"/../gnuplot");
      pathlist.Add(wxFileName(wxStandardPaths::Get().GetExecutablePath()).GetPath()+"/../gnuplot/bin");
      pathlist.AddEnvList(wxT("PATH"));
      gnuplot_binary = pathlist.FindAbsoluteValidPath(m_gnuplotcommand);
      if(gnuplot_binary == wxEmptyString)
        gnuplot_binary = pathlist.FindAbsoluteValidPath(m_gnuplotcommand + wxT(".exe"));
      if(gnuplot_binary == wxEmptyString)
        gnuplot_binary = pathlist.FindAbsoluteValidPath(m_gnuplotcommand + wxT(".exe"));
      if(gnuplot_binary == wxEmptyString)
        gnuplot_binary = m_gnuplotcommand;
    }

In theory if you replace m_gnuplotCommand by Dirstructure::GnuplotDefaultLocation(value) in this place it should be early enough.

...and as a 2nd idea: If wxMaxima knows where gnuplot resides that doesn't help if one wants to instruct maxima to plot a new diagram: Maxima needs to know where gnuplot resides, too, in order to be able to do that. And if maxima knows where gnuplot can be found it will tell us the right gnuplot location. If maxima currently sends us the wrong info: Would you be so kind and tell them that they need to fix this? As a workaround until that is effective we could add a maxima command that corrects the path to gnuplot to m_configCommands. Or does maxima know the correct path to gnuplot because it is run in a shell? And is us not knowing where gnuplot resides a problem only wxMaxima has, because it is a gui application? In this case I would propose that we use the gnuplot maxima tells us to use if we can find it (the user might have customized it), and use the gnuplot we have found, otherwise.

src/Dirstructure.cpp
```diff
@@ -239,6 +242,65 @@ wxString Dirstructure::MaximaDefaultLocation()
...
+#ifdef __WXOSX__
+    // MacPorts:
+    // The MacPorts default binary path /opt/local/bin/ is not in the PATH for applications.
+    // It is added to .profile, but this is only used by shells.
+    // => Add the default MacPorts binary path /opt/local/bin/ to our search path list.
+    //
+    // Homebrew:
+    // Homebrew installs binaries in /usr/local/bin, which is in the PATH by default.
+    //
+    // Application packages including gnuplot:
```
gunterkoenigsmann: If you want to add more wxLogMessages somewhere here: Feel free to do so. Most messages originally were meant as pure eye-candy. But if anything goes completely wrong adding *loads* of log messages has proven helpful for finding out where a problem might reside on a system that doesn't contain a debugger and misbehaves.

src/wxMaxima.cpp
MSoegtropIMC: Fine, i can change back changing m_gnuplotcommand. I did this mostly for performance reasons, but since we don't search paths recursively, it should be fine.

You probably have seen that I also changed the command which (on OSX) injects a command to set the gnuplot path for maxima. This was there before, but used a different heuristic to find gnuplot, which in my view wasn't a good idea.

I didn't find a place where you retrieve the gnuplot command from maxima. Can you point me to it?

gunterkoenigsmann: If maxima informs us about the value of a variable 

     wxMaxima::ReadVariables(wxString &data)

is called. The exact code location is:

            if(name == "gnuplot_command")
            {
              m_gnuplotcommand = value;

Having the gnuplot command configurable by maxima is especially usable on MS Windows: For windows there are two versions of gnuplot that each come with different bugs:

  * gnuplot.exe plots embedded and other plots fine if run from a terminal (and only shows any gnuplot window if the terminal actually opens up in a separate, and complete useless and empty window), but some users want to be able to enable gnuplot's command console. And gnuplot.exe opens the command console in a hidden, and inaccessible window.
  * wgnuplot.exe works just fine: It pops up a window if you want to, is able to open its command console. Bnd even if it plots an embedded plot for the worksheet or an animation frame and therefore doesn't open a window it steals the mouse focus which makes working with plots a nuisance.

I am always thinking of telling maxima to report the value of gnuplot_command every time maxima goes busy, which would allow to change wxMaxima's m_gnuplot_command from the worksheet.

src/Dirstructure.cpp
```diff
@@ -214,7 +217,7 @@ wxString Dirstructure::MaximaDefaultLocation()
     return maximaLocation;
 
   wxLogMessage(wxString::Format(notFound,maximaLocation.utf8_str()));
-  maximaLocation = "/opt/local/bin/maxima";
+  maximaLocation = OSX_MACPORTS_PREFIX "/bin/maxima";
```
gunterkoenigsmann: Perhaps we should rename this to MAXIMA_PREFIX, as fink and homebrew probably need a similar setting... ...and we could document this setting in README.md

src/wxMaxima.cpp
```diff
@@ -3708,9 +3708,9 @@ void wxMaxima::SetupVariables()
   wxString cmd;
 
 #if defined (__WXOSX__)
-  wxString gnuplotbin(wxT("/Applications/Gnuplot.app/Contents/Resources/bin/gnuplot"));
-  if (wxFileExists(gnuplotbin))
-    cmd += wxT("\n:lisp-quiet (setf $gnuplot_command \"") + gnuplotbin + wxT("\")\n");
+  wxString gunplot_binary = Dirstructure::GnuplotDefaultLocation(m_gnuplotcommand);
+  if (wxFileExists(gunplot_binary))
+    cmd += wxT("\n:lisp-quiet (setf $gnuplot_command \"") + gunplot_binary + wxT("\")\n");
```
gunterkoenigsmann: Here I want a wxLogMessage telling the user we have changed the value of gnuplot_command: Else we or the maxima team sooner or later get a but report from a user who tries to change gnuplot_command in the startup file, which for some reason is without effect.

src/wxMaxima.cpp
```diff
@@ -2641,7 +2641,7 @@ void wxMaxima::ReadVariables(wxString &data)
             if(name == "gnuplot_command")
             {
               m_gnuplotcommand = value;
-              wxLogMessage(wxString::Format(_("Gnuplot can be found at %s"),value.utf8_str()));
```
gunterkoenigsmann: This no more is the right place for a message, or the right message. But a wxLogMessage telling which gnuplot binary we have decided upon and why we did this might prove handy, one day: Any decision the program might make might be wrong and it is always good to document such places to make the search for bugs easier.

PR #1249, opened 2020-04-21
Typofix

richardgobeli:


gunterkoenigsmann:
Thanks a lot! Seems I caused a conflict - but that was easy to resolve.

richardgobeli:
I don't know how to resolve the error in verification.  The text is the same. It is just shifted by one word.

yurchor:
Is it possible to unerase purged manual translations (German and Ukrainian)?

Thanks in advance for your answer.

gunterkoenigsmann:
The error in verification was caused by me - and was easy to fix.

Unerasing should be easy as git stores the whole history so one can always go back in time. What did I delete I shouldn't have to?

yurchor:
> What did I delete I shouldn't have to?

Whole German and Ukrainian manual translations.

yurchor:
> How did I cause that?
> Give me a few hours. But I will repair that, too.

No harm was done (I have a backup). Just wanted to know if I should recommit the translation and wanted the German colleagues to be acknowledged. Thanks for your efforts, anyway.

yurchor:
> If you could do so I would be quite thankful.

Done. ;)

gunterkoenigsmann:
I think I have found the bug that allowed me to easily loose translations. Hope I didn't trigger the problem another time whilst trying to fix it.

gunterkoenigsmann:
Wait: You don't mean I have purged the .po file's contents but the translation of the .md and .html files? I think then there is a bug in CMakeLists.txt. Will search for that the day after tomorrow, most probably.

PR #1250, opened 2020-04-21
Typofix

richardgobeli:
Changed an to a.


gunterkoenigsmann:
Seems like this time the error, too, wasn't in the code (which wouldn't be a big problem: If it is broken it can be repaired), but in AppVeyor or in any of the services AppVeyor uses.
How appveyor internally works I don't know. But I know that travis, which is a similar service rents a work queue manager, one virtual machine creation service, one service that runs the virtual machines, load balancers, databases - all from different firms and often builds fail due to one of these services having a small outage.

richardgobeli:
Thanks,  I did not know If I did something wrong.   The first one today, I though the two files were the same but split per line different.

PR #1257, opened 2020-04-24
New upstream master

KubaO:
Since all the font cache changes should be undone, pull this branch, then check out master, `reset --hard` the master to this branch, then force push to replace the branch in GitHub. A bit of history rewriting :)

### DIFF
--- a/info/wxmaxima.md
+++ b/info/wxmaxima.md
@@ -288,7 +288,7 @@ You can be use `.mac` files for writing own library of macros. But since they do
 
 ### .wxm
 
-.wxm files contains the worksheet except _Maxima_'s output. On Maxima versions >5.38 they can be read using _Maxima_'s `load()` function just as .mac files can be. With this plain-text format it sometimes is unavoidable that worksheets that use new features are not downwards-compatible with older versions of _wxMaxima_.
+.wxm files contain the worksheet except _Maxima_'s output. On Maxima versions >5.38 they can be read using _Maxima_'s `load()` function just as .mac files can be. With this plain-text format it sometimes is unavoidable that worksheets that use new features are not downwards-compatible with older versions of _wxMaxima_.
 
 ### .wxmx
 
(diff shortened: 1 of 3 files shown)


gunterkoenigsmann:
I believe that hard resetting the master breaks updating the repo for everyone else (requires a force update). But I believe I have an idea on what to do.

gunterkoenigsmann:
Wait... ...If I pull your branch I get a months old wxMaxima version What am I doing wrong?

gunterkoenigsmann:
Alternatively I can "git revert" any (or all) of my recent commits - or you have the rights to do this if that helps.

As I have stated I believe you cannot change the history of the master branch without causing a surprisingly big number of problems for all that are currently on the master branch. But if you want to do anything else: I guess you'll have received a message about write rights from github.

KubaO:
In the future: please don't blindly apply any pull requests. They have to be tested and discussed first. Also, I can't do anything since I don't have write access. I suggest you don't revert since that just makes a mess of the commit history. This "diversion" has to be completely purged :)

KubaO:
> Wait... ...If I pull your branch I get a months old wxMaxima version What am I doing wrong?

I have no clue, sorry. The branch is based on the master from yesterday...

gunterkoenigsmann:
You tell me how to run my projects?

Hmmm... 

My normal strategy is to assume that everything that builds on all available CIs (appveyor, at this time, since travis was broken) and comes from a developer who can be trusted to provide cool magic everything has been tested before filing a PR and if it isn't - every step forward is good and if a detail is missing that can be un-broken. Also I still am convinced that the font cache wasn't a first step into the right direction. Especially as as you have said wxMaxima is much slower than I believe it should be. Your new idea I believe, though, to be even better.

On the other hand I have to admit that my strategy is non-standard. But nothing is lost if we don't break "git update" for the master branch (if that could not be avoided that wouldn't be a catastrophe, neither, probably) => feel free to use "git revert" on any commit you want (or tell me if to only revert my commits or also merging the pull request this way). Or merge everything you want with the contents of your branch by manually overwriting the files or use "git merge" in a cool way to archive this goal. Or tell me how to pull a branch I can overwrite the contents of my branch with (save the new translation, of course) - and we'll be another step towards a better tool.

Normally there always is a way.

gunterkoenigsmann:
You cannot rewrite the history of a branch others are working with, too nor force-push to these branches if you aren't the only one using them - except if you can make sure that no-one currently has checked out a commit that will disappear when you change the history. That isn't a problem with private branches. This allows some projects to require the commit history of private branches to be clean before they are merged. Doing all of the development in branches even enables keeping the master branch clean in most cases.

I personally don't care about a clean commit history. I guess I should start to do so for my own commits, now that there are many developers in the project, though. ...but I don't want to cause others hard-to-resolve problems by force-pushing. But I still don't understand why I get the wrong result if I try to checkout your branch...

gunterkoenigsmann:
Tried to checkout the branch another way - but did get a 404 error, this time => I guess the branch is deleted now. Or id I somehow all the time try to checkout the wrong branch? That would explain a lot.

Is there still a way to still get your files? A chance of Speeding up the programs (and any chance of having a capable developer who has ideas on how to do so) is important.

PR #1258, opened 2020-04-27
Fix the accessibility method signatures so that they actually implement wxAccessible API.

KubaO:
This gives us a chance of being somewhat accessible at least.

### DIFF
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -1320,8 +1320,8 @@ wxAccStatus Cell::GetDescription(int childId, wxString *description)
   }
   else
   {
-    Cell *cell = NULL;
-    if(GetChild(childId,&cell) == wxACC_OK)
+    wxAccessible *cell = NULL;
+    if(GetChild(childId, &cell) == wxACC_OK)
     {
       if(cell != NULL)
         return cell->GetDescription(0, description);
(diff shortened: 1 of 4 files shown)


gunterkoenigsmann:
Wow... ...I cannot test this (no windows computer ( but If you tell me I can merge this: There are blind maxima users that still hope that wxMaxima one day will be accessible to them and If we were able to help them that would be really good news.

KubaO:
This is ready to merge. It’s a fix for wrong method signatures. With C++11, the override keyword prevents such bugs, but this was older code. I’m also not sure if wxWidgets doesn’t have accessibility support under gtk and/or Mac - Kuba

gunterkoenigsmann:
C++11 has brought great improvements for us.

As trusty tahr nears the end of non-paid support I am currently evaluating if we can dare to fast-forward to C++14 as you told you cannot go farther back in time.

KubaO:
Personally, I’d love C++17 since it got string_view, and that’s needed to fix the horrible string copying in the parser and tokenizer (among others). Lots of time is spent there. The unoptimized debug builds really amplify all that and make it easy to spot: open a big document, then randomly pause in the debugger and see what is being done. Other than the fonts, it’s almost always wxString copying or reallocations.

I have a copy/allocation reduction branch on GitHub and there’s lots to be done there. That’s the major problem with wxWidgets: it’s getting long in the tooth, technology wise. The font caching work I’m doing also would ideally go into wxWidgets itself.

But I can live with C++14 for sure. It may be a problem for older MacOs versions, but I imagine there can be workarounds as long as those are only the libraries that are concerned, not, say, fancy constexpr etc.

gunterkoenigsmann:
C++17 has many cool features. But it is quite hard for some users to switch to a new enough compiler => I guess we'll have to wait at least 3 years before switching...

PR #1260, opened 2020-05-01
Feature/node refactor 1

KubaO:
This replaces inner node lists with vectors, and uses make_shared to create the instances.

There are many other uses of std::shared_ptr(new Class(...)) that can be replaced, but it looks like all of the shared pointers can be unique pointers anyway, so that's what the future PRs will be about.

### DIFF
--- a/src/AbsCell.cpp
+++ b/src/AbsCell.cpp
@@ -36,7 +36,7 @@ AbsCell::AbsCell(Cell *parent, Configuration **config, CellPointers *cellPointer
   m_close(new TextCell(parent, config, cellPointers, wxT(")"))),
   m_last(NULL)
 {
-  m_open->DontEscapeOpeningParenthesis();
+  static_cast<TextCell&>(*m_open).DontEscapeOpeningParenthesis();
   m_open->SetStyle(TS_FUNCTION);
 }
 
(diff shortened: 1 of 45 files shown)


gunterkoenigsmann:
That looks like a really valuable contribution - even if the "lists are slower than vectors" normally triggers a long discussion as lists are half-way fast if you always iterate through all of the list

Would you mind adding another commit to your branch that sorts out the differences between our compilers? In my experience that automatically updates the pull pequest.
~~~~
/home/gunter/src/wxmaxima/src/Cell.cpp:219:15: warning: Consider using std::accumulate algorithm instead of a raw loop. [useStlAlgorithm]
        cells += cell->CellsInListRecursive();

/home/gunter/src/wxmaxima/src/SubSupCell.cpp: In member function ‘virtual wxString SubSupCell::ToString()’:
/home/gunter/src/wxmaxima/src/SubSupCell.cpp:290:28: error: ‘as_const’ is not a member of ‘std’; did you mean ‘is_const’?
  290 |     for (auto &cell : std::as_const(m_innerCellList))
      |                            ^~~~~~~~
      |                            is_const
/home/gunter/src/wxmaxima/src/SubSupCell.cpp: In member function ‘virtual wxString SubSupCell::ToMatlab()’:
/home/gunter/src/wxmaxima/src/SubSupCell.cpp:318:28: error: ‘as_const’ is not a member of ‘std’; did you mean ‘is_const’?
  318 |     for (auto &cell : std::as_const(m_innerCellList))
      |                            ^~~~~~~~
      |                            is_const
~~~~
Thanks a lot in advance,
and kind regards,

         Gunter.

PR #1261, opened 2020-05-01
Reduce data structure overhead. Fix a broken regex.

KubaO:


gunterkoenigsmann:
I have just fixed a monster bug: A bug that causes parts of equations not to be displayed.
Hoped that such bugs were eliminated with the C++11 transition ;-(

Anyway: I wanted to inform you that as soon as you merge the new and faster font cache I want to make a new release: Loosing user data is bad. But the new font cache feels like being a killer feature to me. Hope this is a good decision.

KubaO:
Sorry for the bug!!

The new font cache extends the Style class to be used instead of wxFontInfo, and adds fully hierarchical styles where a style can have unset values that propagate to the base style. That gets rid of the logic that treats some styles specially that’s all over the place. Instead of GetDefaultFontSize etc in configuration, the TS_DEFAULT style carries that data, and there’s also a TS_MATH_DEFAULT and TS_CODE_DEFAULT as base styles for math and code, respectively.

It’s in the bug fix phase, where I have to ensure it’s compatible with old wxConfig values, and that the defaults end up being the same. Basically, users of fonts do this: configuration->GetStyle(TS_FOO).GetFontAt(fontsize) to get a style’s font at a custom size, or GetFont() to get the font at size set by the style (taking configuration’s scaling into account). So all the code that currently has to tweak fonts all over the place is gone. That will take still a week or two to get shaken down and then I’ll unleash it on you to find bugs I’d surely miss. Setting up tests for that will be a whole another chunk of work that I’ll have to do separately.

I still have lots of pointer and data structure fixing - the MarkAsDeleted code will be completely gone, weak_pointer will do the job of tracking object lifetime, I have also lots of string copying in the markdown parser and in the tokenizer that I want all gone. All of that stuff shows on profiles and makes debug builds unusable with long documents. It doesn’t help that wxString is pretty bad.

My goal for the release is to have an 80 page benchmark document open in unoptimized debug build in a few seconds. I’ll replace all text and variable names with “Lorem Ipsum” and add it to the repository as a test case. 

Some very future step for me may be to fork wxWidgets or to even refactor the wxMaxima code so that the GUI areas would be separate and make it easier to port to any other framework (what LyX people had to do long ago).

There are lots of wxWidgets inefficiencies that could be fixed by porting the code to C++11, and getting rid of legacy compatibility stuff we don’t use in wxMaxima - stuff that won’t be accepted into wxWidgets upstream since they need to be compatible with old code and old compilers (they still need to compile with VS6 it seems). But that’s not too critical at the moment - tweaking wxWidgets only makes sense after there’ll be no more benefits to be had from wxMaxima improvements.

gunterkoenigsmann:
I found another thing that reduces the performance on linux - but for that I have no good idea on how to fix that: DC::SetFont and changing the foreground and background color the DC uses seem to be slow on linux. Often we call these functions without changing the font and colors to a new value. Perhaps *Cell::Draw shouldn't set the font and color but GroupCell:Draw() should do that for the cells keeping track of the changes. But that doesn't fell like an easy-to-understand method for others who look at the code...

gunterkoenigsmann:
If you get rid of MarkAsDeleted by using weak_ptr you are a real wizard! I tried to do so once, but never got it to work.

And the big bug I fixed wasn't caused by any of your work: It was there since at least 2013.

---

Vadim, the wxWidgets maintainer a few weeks a got if they were to switch modern features he would make a big jump to C++11. But from there it would be a long time before every Linux distri switches to the new version and therefore we can make a hard dependency on it without causing big problems for many users...
...and I want wxWidgets to support svg. Which they nearly started a gsoc about.

PR #1268, opened 2020-05-08
Feature/fix char cache

KubaO:
The cache was caching a negative result while returning a positive result. There were several allocations when hitting a hot cache. Since the cache is tiny, a vector works better than any other data structure.


gunterkoenigsmann:
That is nice code!

PR #1269, opened 2020-05-09
Feature/internal cell iterators

KubaO:
This gets rid of GetInternalCells and the associated allocations. Instead, `Cell::InternalBegin()` and `Cell::InternalEnd()` provide iterators over the internal cells. Iterating over internal cells is a hot path and does not allocate.

The accessibility code is reworked to use these iterators, getting rid of repeated cell counting. The accessible cell ids are generated on the fly while the cells are iterated, and there's no need to count cells ahead of time.

### DIFF
--- a/src/AbsCell.cpp
+++ b/src/AbsCell.cpp
@@ -61,18 +61,6 @@ AbsCell::~AbsCell()
   MarkAsDeleted();
 }
 
-Cell::InnerCells AbsCell::GetInnerCells() const
-{
-  InnerCells innerCells;
-  if(m_innerCell)
-    innerCells.push_back(m_innerCell);
-  if(m_open)
-    innerCells.push_back(m_open);
-  if(m_close)
-    innerCells.push_back(m_close);
-  return innerCells;
-}
-
 void AbsCell::SetInner(Cell *inner)
 {
   if (inner == NULL)
--- a/src/AbsCell.h
+++ b/src/AbsCell.h
@@ -63,7 +63,8 @@ public:
   //! This class can be derived from wxAccessible which has no copy constructor
   AbsCell &operator=(const AbsCell&) = delete;
 
-  InnerCells GetInnerCells() const override;
+  InnerCellIterator InnerBegin() const override { return &m_innerCell; }
+  InnerCellIterator InnerEnd() const override { return &m_close+1; }
 
   void SetInner(Cell *inner);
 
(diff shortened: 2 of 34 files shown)


gunterkoenigsmann:
I am afraid I need a little bit of help with this one: It works perfectly and looks quite nice to me. But cppcheck complains a few times and I am not sure how to get rid of that:
````
/home/gunter/src/wxmaxima/src/Cell.h:1048:5: warning: Class 'InnerCellIterator' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
    InnerCellIterator(const std::shared_ptr<Cell> *p) : ptr(p) {}
    ^
````
Normally putting an `explicit` in front of line 1048 would resolve that. But this time gcc no more seems to find the constructor if I do that

````
/home/gunter/src/wxmaxima/src/Configuration.cpp:766:0: warning: Consider using std::find_if algorithm instead of a raw loop. [useStlAlgorithm]
    if (ex.chars == chars)
^
````
I like this piece of code and sometimes cppcheck fires on constructs that are completely fine and cannot be further optimized. Is this the case here? If yes I will try to mute this warning.

gunterkoenigsmann:
Hope my changes to the math parser were a good idea...

gunterkoenigsmann:
At least the math parser looks cleaner now => Started a new branch named GroupcellParserSimplification in which I try to further clean up the math parser code.

KubaO:
Haven’t looked at it yet but any change in that area will probably be good.

The markdown parsing is also totally broken for export to Tex/html. That one should probably use an external library (header only), and should be merged with the one used for syntax highlight and real-time substitutions. But it’s not very high priority for me. I took a stab and gave up - it was too broken, there’s no saving that code at all.

gunterkoenigsmann:
At least I now understand how wxmx parsing is done. But I guess you would have added much magic like the elegant `for` construct cppcheck used to complain about (but no more does in their newest release).

---

The problem with external libraries is that it is hard to find users who manage to install them on their computer. We depend on wxWidgets - which is hard enough to install on platforms except linux, cmake (many users have big problems compiling it) and a C++ compiler which some people deem to be more dangerous than wild horses. Adding more dependencies causes a big lot of trouble. Also some of markdown is indistinguishable from mathematical notation which is why supporting more than a small subset of markdown might not be a good idea.

But you are right: My makeshift MarkDown parsing code is historically grown and replacing it would greatly improve the overall code quality.

PR #1275, opened 2020-05-12
Feature/refactor clipboard and export

### LINKED ISSUE #1274: EMF Copy to Clipboard doesn't work on Windows
I'm copying EMF to Windows clipboard, either using Edit->Copy as EMF, or just doing Ctrl-C with EMF selected in the configuration dialog. The clipboard gets some EMF data posted to it (there's `image/x-emf` listed as a data format), but:

1. The Preview looks empty when viewing using Free Clipboard Viewer 3.0.

2. No applications accepts the EMF for pasting. Bitmap and text formats work OK.

This is both on the master HEAD as well as 20.04.0 release.

I'll try and investigate. The temporary EMF files are valid and both MS Paint and LibreOffice Draw open them fine. It seems that the string clipboard format `image/x-emf` is not recognized on Windows.

OK, I figured it out. The problem is that `wxEnhMetaFileDataObject` has to be used to represent the enhanced metafiles, instead of the custom data object used thus far. But it's not a `wxSimpleDataObject`, and thus cannot be aggregated using the `wxCompositeDataObject`. A custom aggregate class needs to be devised that can compose arbitrary `wxDataObject`s, not merely `wxSimpleDataObject`s.

KubaO:
1. Factor out duplicate code in the SVGout, EMFout and BitmapOut classes.
2. Implement `CompositeDataObject` to replace the silly `wxDataObjectComposite`, allowing compositing complex data objects (those that have more than one format), such as EMF files.
3. Fix EMF format not making it to the clipboard on Windows. Fixes #1274.

Note that the `CompositeDataObject` class can be used as a superior replacement to `wxDataObjectComposite` on all platforms, but thus far I've only substituted it when the EMF output is enabled on Windows builds. I'm sure we'll run into a similar problem on other platforms sooner or later, and can switch to it on all platforms. But so far I wanted to minimize the changes to only those that were necessary - i.e. on builds where EMF output is enabled.

### DIFF
--- a/src/BitmapOut.cpp
+++ b/src/BitmapOut.cpp
@@ -25,270 +25,87 @@
  */
 
 #include "BitmapOut.h"
-#include "Configuration.h"
-
-#include <wx/config.h>
+#include "Cell.h"
 #include <wx/clipbrd.h>
 
 #define BM_FULL_WIDTH 1000
 
-BitmapOut::BitmapOut(Configuration **configuration, int scale):
-  m_dc(new wxMemoryDC())
+BitmapOut::BitmapOut(Configuration **configuration, int scale) :
+    m_cmn(configuration, BM_FULL_WIDTH, scale)
 {
-  m_scale = scale;
+  m_cmn.SetSize({10, 10});
+  m_bmp.CreateScaled(10, 10, 24, scale);
+  m_dc.SelectObject(m_bmp);
+  m_dc.SetUserScale(scale, scale);
+  m_dc.SetPen(wxNullPen);
+  m_cmn.SetRecalculationContext(m_dc);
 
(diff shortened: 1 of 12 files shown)


gunterkoenigsmann:
Another question:
````
  if (!line.empty())
  {
    GroupCell *cell;
    appendCell((cell = new GroupCell(config, GC_TYPE_CODE, cellPointers, line), last));
  }
````
What if line is still empty when encountering the "if"? What is the sense of the ",last"? And "cell" happens to be the name of a variable in the outer loop. Can I rename it?

KubaO:
If the line is still empty then no cell should be created I think. I need to look again at this.

PR #1279, opened 2020-05-13
Fix an uninitialized member.

KubaO:
This pointer was not initialized and could be potentially used uninitialized.


gunterkoenigsmann:
I was completely sure that either cppcheck or the c compiler would automatically find such a thing...

KubaO:
Well it did :) Most recent one at least.

Issue #36, opened 2011-04-20
Cannot connect to maxima


andrejv:
Converted from [SourceForge issue 1573741](http://sourceforge.net/support/tracker.php?aid=1573741), submitted by rexdieter

See also
http://bugzilla.redhat.com/209992

On second and subsequent runs, an invalid entry in
~/.wxMaxima of
maxima=1
causes subsequent evocations to fail.

Patch forthcoming.

Issue #35, opened 2011-04-20
opens infinite number of "tip of day" dialogs


andrejv:
Converted from [SourceForge issue 1561644](http://sourceforge.net/support/tracker.php?aid=1561644), submitted by lindi2

I recently recommended wxmaxima to a windows user and
got a reply that

http://prdownloads.sourceforge.net/wxmaxima/wxMaxima-0.6.5.exe?download

opens infinite number of "tip of day" dialogs. I don't
unfortunately have much details since I don't
personally use windows anywhere. I only know he has
Windows XP. If you think there is some specific error
case that could cause the problem I can try to ask for
more info. If you don't see any way this could happen
please consider this all to be a rumor and close this
bug with INVALID ;-)

PR #1282, opened 2020-05-15
Copy less in the wxm parser

KubaO:
This is minor but helps - the line array the parser works on is not changed nor copied.

### DIFF
--- a/src/GroupCell.cpp
+++ b/src/GroupCell.cpp
@@ -39,7 +39,7 @@
 #include "BitmapOut.h"
 #include "list"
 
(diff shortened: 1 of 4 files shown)


gunterkoenigsmann:
Travis claims that after that change image cells cannot be loaded any more:
````
35/85 Test #35: slideshowCells ..............................   Passed    0.61 sec

      Start 36: imageCells

36/85 Test #36: imageCells ..................................***Exception: Child aborted  0.46 sec
````
Today I won't have the time to look if that claim is correct, unfortunately.

PR #1284, opened 2020-05-16
Feature/refactor wxm format

KubaO:
Factor out WXM read/write formatting code. This allows the conversion between cells and WXM format to reuse the same header strings, for consistency/DRY/single source of truth.

Factor out the rest of the .wxm and .mac file parsing. he wxMaxima.cpp file is big enough already.

The changes mostly revolve around removing repetition.


gunterkoenigsmann:
That is a thing I always wanted to do and never dared to do as this is a change that is big enough to be guaranteed to introduce new bugs. Our automatic tests are convinced that this happened only in one occasion that was trivial to fix => Good work.

gunterkoenigsmann:
If you want to work in the main repo (in which commits automatically trigger all of the available tests): Feel free to do so. As long as there is no history rewriting in already-pushed commits in the master branch I won't mind force-pushes or experimental branches there.

KubaO:
I wouldn’t mind working on the main repo. I always work in feature branches and all sorts of “violent” things happen within those, and then do a single final merge to main branch. What I like to do though is to keep the few individual commits in those features without squashing as they make later blame/inspect operations easier (if something breaks and you want to know what broke and why). The main branch sees single merge commits, but all the steps are available for those who need them.

Most features I do consist of just a few commits -often 2 or 3, but sometimes there is more if it’s a major change.

There’s lots of repetition in the code and I love getting rid of duplication while also improving performance - wxMaxima is a long-lived code base and it’s natural that it accumulated some “crust”. It certainly has a potential to be a lot more, but a lot of grunt work has to be done to allow that, and to attract other developers.

gunterkoenigsmann:
Tried to send you another invite for write access to the main repo.

PR #1314, opened 2020-06-22
typo in toleranceCalculations Example

mestrode:
The tolerances of the second voltage divider are linked to each other, think this is not intended.


gunterkoenigsmann:
The question is if could describe my intention better in the example's text: I wanted to demonstrate the case that two tolerances are linked - and one example of that I frequently have to explain is:

 * If you have two excellent resistors, but each of them has its own temperature coefficient the resisting voltage divider may have a noticeable temperature coefficient.
 * But if you have two resistors whose absolute value isn't known, while the ration between them is known and isn't temperature-dependent you will have an excellent voltage divider instead.

For the latter case I need the tolerances to be linked.

PR #1392, opened 2020-08-31
make wxxml output tolerant of malformed expressions

robert-dodier:
Make wxxml output tolerant of malformed expressions by punting to WXXML-FUNCTION. This is the same behavior as the built-in Maxima 1-d and 2-d display.

This PR contains 4 commits; only commit 5ec5bb6 is relevant. Of the other 3, one is a merge commit, and the other two are an attempted bug fix and its reversion (so net zero change).


gunterkoenigsmann:
Wow! Thanks a lot!

Currently clicking on various worksheet objects is broken. But with a bit of luck I can make a fresh release within a week.

KubaO:
That’s great! In the future, please edit the history so that “net no change” commits aren’t present unless they represent an important attempt that should remain in the project history.

robert-dodier:
Hi Kuba, yes, you're right. I was in a bit of a hurry, but for sure next time I will rebase or squash commits or whatever is necessary to filter out the irrelevant stuff. Thanks for your help, and all the best.

PR #1409, opened 2020-10-06
Feature/koen gu/clean up config

gunterkoenigsmann:


KubaO:
Brilliant! Thank you!

PR #1442, opened 2020-11-25
Update Russian translation

OlesyaGerasimenko:


locales/wxMaxima/ru.po
```diff
@@ -8643,7 +8637,7 @@ msgid "We can introduce (\"substitute\") them into other equations:\n"
 msgstr "Можно вставить («подставить») их в другие уравнения:\n"
 
 #: ../../examples/examples_gettext.h:1788
-#, fuzzy, c-format
+#, c-format
```
yurchor, line 8640: This will not work. You should replace it with "no-c-format".

Please try to run "msgfmt -vc -o /dev/null ru.po" to see the error messages.

yurchor (changes requested):

locales/wxMaxima/ru.po
gunterkoenigsmann: The c-format is autogenerated by msgmerge. So I am unsure if it might automatically reappear at some time.

yurchor: > The c-format is autogenerated by msgmerge. So I am unsure if it might automatically reappear at some time.

Sure. But without removing c-format from two broken strings you cannot compile wxMaxima.

yurchor: What if we merge this PR, then I will fix it this time, but the next time the file will be checked by the committer before PR creation?

gunterkoenigsmann: With merging and fixing I am fine. But instead of adding another task for the translator I would like to resolve the underlying problem.

My guess is that the problem arises only within strings that contain a "%" => Will try to find a way to hinder gettext from using c-format in this file.

gunterkoenigsmann: Thanks a lot! Am currently trying to rename the `.h` file that caused these `c-format` markers to `.list` hoping that this will trick cmake into believing we don't use c code here.

PR #1468, opened 2021-01-05
Re-enable one threadtest

gunterkoenigsmann:


LiuRong0815:
@gunterkoenigsmann Hi, Gunter, I was updating the Chinese manual as before, but the git told me that I "Merge branch 'master' of github.com:wxMaxima-developers/wxmaxima". Seems like file ".github/workflows/compile_ubuntu.yml" has been modified. I am not familiar with git, so if I did something wrong, hope you could fix it, thanks.

gunterkoenigsmann:
Mi interpretation is that git just wanted to document that your change wasn't made to the newest version and git just wanted to document that somewhere => for this commit git officially documented

 - after which change the branch on your computer deviated from the one on the server and 
 - that it merged your branch into the one that can be found on the server after my commit.

There are projects that in such a case want contributers to use git's "rebase" feature: rebase can rewrite your local history to give everyone the illusion that contributions were made in sequential order instead of in parallel. But using rebase is complicated, can go seriously wrong (it is a command that can actually use part of the history including the part of the history other users still are working on) and I don't believe that just because one can fake an aspect of history one is obliged to do so => if you ask me: If two people work in parallel that is documented by a merge commit and merge commits are fine.

----

Generally the translation system is very well-behaved: If I happen to change strings in a file you have already worked on and my work conflicts with yours just undo my changes and commit yours. In that case the system will at the next opportunity re-try to merge the new strings into your file.

If something else goes wrong: git will know all the changes that might have gotten lost, I have configured git to use meld for graphically showing any change, if needed and normally a "git diff" or "git mergetool" allows to clean up anything that can go wrong. Kuba Ober uses a even more advanced tool for the same purpose but currently I don't seem to remember its name...

LiuRong0815:
Thanks for the reply. I always pulled the repo with --no-rebase option just before I am going to push the translations, which make the other files in my local repo up-to-date---- I think. Looking forward to see the new version on the Manjaro, and I will check the translations again after the update.

gunterkoenigsmann:
I have configured my git to use that option, too. A few git versions ago it was standard. And pulling does exactly what you expect: It pulls in all new changes.

PR #1476, opened 2021-01-23
 Fix double release of OMP locks in several places

### LINKED ISSUE #1474: Latest tag crashes on Mac when opening a file containing images when OpenMP is enabled
I wanted to create a new MacPorts release of wxMaxima.

The latest tag (Version-20.12.2) generally looks fine - e.g. the menue issue seems to be fixed. The only issue I have is that it crashes when opening files containing images when OpenMP is enabled. It works fine without OpenMP.

I am still figuring out how to debug this - something seems to have changed in XCode since last year. A symbols only call stack of the crash is:

```
  * frame #0: 0x0000000101e4fbfc libomp.dylib`__kmp_release_queuing_lock + 166
    frame #1: 0x0000000101e52b38 libomp.dylib`__kmp_release_queuing_lock(kmp_user_lock*, int) + 9
    frame #2: 0x000000010027e120 wxmaxima`Image::LoadImage_Backgroundtask(wxString, std::__1::shared_ptr<wxFileSystem>, bool) + 7600
    frame #3: 0x0000000100261b66 wxmaxima`Image::LoadImage(wxString, std::__1::shared_ptr<wxFileSystem>, bool) + 3634
    frame #4: 0x000000010026058e wxmaxima`Image::Image(Configuration**, wxString, std::__1::shared_ptr<wxFileSystem>, bool) + 1602
    frame #5: 0x00000001008247a0 wxmaxima`ImgCell::ImgCell(GroupCell*, Configuration**, wxString const&, std::__1::shared_ptr<wxFileSystem
```

As far as I can tell it is a write to invalid memory. I compiled it with Address Sanitizer, but this doesn't show an issue since libopenmp is not instrumented.

Also I wonder why OpemMP is recommended? I always worked with the OpenMP version and testing now the non OpenMP version doesn't look that different.

MSoegtropIMC:
This PR fixes a bug that OMP locks are released twice, once via a helper class and once explicitly. This leads to crashes on Mac if OMP is enabled.

This fixes #1474

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,4 @@ deploy_rsa
 .ninja_log
 build.ninja
 rules.ninja
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,5 @@ deploy_rsa
 build.ninja
 rules.ninja
 .DS_Store
+.vscode/
+wxmaxima.code-workspace
--- a/src/Image.cpp
+++ b/src/Image.cpp
@@ -803,9 +803,6 @@ wxBitmap Image::GetBitmap(double scale)
   }
   else
     m_scaledBitmap = wxBitmap(1,1);
-  #ifdef HAVE_OMP_HEADER
-  omp_unset_lock(&m_gnuplotLock);
-  #endif
   return m_scaledBitmap;
 }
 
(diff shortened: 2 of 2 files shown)


gunterkoenigsmann:
Thanks a lot!

PR #1496, opened 2021-02-23
Fix and adjust preferences dialog  size

MSoegtropIMC:
This fixes the issue that on macOS the preferences panel size was set to very small after changing the panel.

This also slightly adjusts the panel size, so that all panels fit on Mac (I made it a bit wider and a bit less tall, from 600x800 to 640x720). It needs to be checked if this also works nicely on other systems, but anyway scroll bars are enabled in case the panels are too large on some systems.

@gunterkoenigsmann : FYI.

### DIFF
--- a/src/ConfigDialogue.cpp
+++ b/src/ConfigDialogue.cpp
@@ -263,14 +263,10 @@ m_maximaEnvDoc["GCL_MULTIPROCESS_MEMORY_POOL"] = _("If maxima was compiled by GC
      notebookTab = 0;
   m_notebook->SetSelection(notebookTab);
 
-  // Allow the property dialogue sheets to scroll, if they don't fit
-  // on the screen.
+  // Allow the property dialogue sheets to scroll, if they don't fit on the screen.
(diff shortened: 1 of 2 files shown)


MSoegtropIMC:
@gunterkoenigsmann : can you please check if the slightly modified sizes look nice on Windows, possibly enlarge them and then make a tag? I would do a new MacPorts release then.

MSoegtropIMC:
P.S.: The SetInitialSize has no effect on Mac if scroll bars are enabled, so I removed it.

P.P.S.: I don't understand the CI issue - it seems to be unrelated to the changes I did.

gunterkoenigsmann:
I don't own a windows PC. But your code looks good.

MSoegtropIMC:
@gunterkoenigsmann : ah ok - somehow I thought you are on Windows. That's fine, I can check it on Windows as well. Can I download a nightly build Windows installer somewhere?

MSoegtropIMC:
P.S.: what are your plans for the next release?

PR #1576, opened 2021-08-23
Make GetDiffPart work with different cell types

### LINKED ISSUE #1575: diff is not copied correctly
When the variable of differentiation is something simple (like x) that can be represented by a TextCell, one can copy the output and get the correct maxima command:

```
diff(f(x), x);  /* input */
'diff(f(x),x,1) /* copied output */
```

But if the variable w. r. t. which differentiation takes place is represented by something like SubCell or FunCell, it doesn't get copied:

```
diff(f(x[1]), x[1], 1);  /* input */
'diff(f(x[1]))           /* copied output */
```
```
'diff(f(x), g(x));  /* input */
'diff(f(x))         /* copied output */
```

This was originally mentioned in [this StackOverflow question](https://stackoverflow.com/q/68290774/14773728).

gcem:
This PR closes #1575.

First commit fixes the `ToString` (copy) for the inputs below, by adding a default `GetDiffPart` function that should work with many cell types:

```
diff(f(x[1]), x[1], 1);
```
```
'diff(f(x), g(x));
```
```
'diff(f(x),g(x),2);
```

Second commit fixes it for this input, where the variable of differentiation has a subscript and the order of differentiation is higher than one (therefore `SubSupCell` is used):

```
'diff(f(x[1]),x[1],3);
```

---

One problem that remains is that `'diff(f(x), x^2, 1)` results in the operator `d/(dx^2)`, and wxMaxima takes the exponent 2 in the denominator as the order of differentiation. But I think this should be a separate issue.

### DIFF
--- a/src/cells/Cell.cpp
+++ b/src/cells/Cell.cpp
@@ -942,7 +942,7 @@ wxString Cell::ListToXML() const
 }
 
 /***
- * Get the part for diff tag support - only ExpTag overvrides this.
+ * Get the part for diff tag support - only ExptCell and SubSupCell override this.
  */
 wxString Cell::GetDiffPart() const
 {
--- a/src/cells/Cell.cpp
+++ b/src/cells/Cell.cpp
@@ -946,7 +946,11 @@ wxString Cell::ListToXML() const
  */
 wxString Cell::GetDiffPart() const
 {
-  return {};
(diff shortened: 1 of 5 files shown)


gunterkoenigsmann:
Thanks a lot!

PR #1585, opened 2021-09-22
Subscript fix

### LINKED ISSUE #1584: Underscores not converted to subscripts
wxMaxima 21.x doesn't convert underscores to subscripts properly. In the below example t_ONmax and t_OFFmax should be displayed with subscript.
grafik
Couldn't detect a scheme so far, which underscores are converted to subscript and which ones not.
Older versions like 17.05.0 displays the same formula (and of course all others as well) correctly:
grafik

wxMaxima version: 21.02.0_MSW
using wxWidgets version: wxWidgets 3.1.4
Maxima version: 5.45.1
Maxima build date: 2021-06-21 17:43:28
Host type: x86_64-w64-mingw32
System type: Win32 6.2.9200 X86-64
Lisp implementation type: SBCL
Lisp implementation version: 2.0.0

gcem:
This may help resolve #1584.

I also saw another problem with the subscripts while looking into this:

grafik

We have an option to convert all underscores to subscripts. The other option doesn't convert when neither part of the expression is a single character or number (e.g. `aa_aa`). Currently we always do the latter, regardless of the option's value.

I have also fixed this in this PR. But now, functions loaded with `load` function are getting subscripted:

grafik

So if we will go with this change, it may be better to change the option's default to "Integers and Single Letters".

Also, I got an error similar to "variable stack overflow" while testing this. I didn't see it later but I need to make sure it wasn't caused by my changes.


gcem:
Now I see another problem with this PR. I think it needs more work.

grafik

PR #1595, opened 2021-10-14
Correctly set the document saved state when the size of an image is changed.

wiesiwies:
Just setting m_fileSaved to false does not work, because it gets overwritten at
other locations.


wiesiwies:
BTW: There may be more occasions of this problem in other code parts. I've seen it at least once today but couldn't remember the trigger when I realized it happening.

PR #1597, opened 2021-10-17
Enable reloading and changing images inserted via "Cell" -> "Insert Image..."

wiesiwies:
This branch adds "Reload" and "Change Image" items to the image cell context menu:
reloading_enabled

- "Reload" reloads the file from disk using the original image filename and path. This is useful if the image was edited externally and it needs to be updated in the worksheet.
- "Change Image" opens a file dialog where an alternative file for replacing the current image can be selected.

The original image filenames and paths are stored in the worksheet file to keep the reload capability after re-opening the worksheet. Since users may not want to leak filenames and paths in their worksheets, a configuration checkbox for disabling storing the filenames has also been implemented.
cfg_dialog
The option defaults to off (not storing in the worksheet file) which prevents `ImgCell::ToXML` from outputting the filenames.

If a filenames is not found, the reload function is disabled
reloading_disabled
Changing the image is, of course, still possible.

### DIFF
--- a/src/ConfigDialogue.cpp
+++ b/src/ConfigDialogue.cpp
@@ -459,6 +459,7 @@ void ConfigDialogue::SetCheckboxValues()
   else
     m_ctrlEnterEvaluates->SetValue(true);
   m_numpadEnterEvaluates->SetValue(configuration->NumpadEnterEvaluates());
+  m_saveImgFileName->SetValue(configuration->SaveImgFileName());
   m_saveUntitled->SetValue(configuration->SaveUntitled());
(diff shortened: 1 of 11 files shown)


gunterkoenigsmann:
Under which name should I add you ti the authors list?

wiesiwies:
> Under which name should I add you ti the authors list?

Thanks for asking, it is Thomas Wiesner.
(I was actually unaware, that somewhere in the push/pull request/merge process the identity data of the local repo gets replaced with the github account settings. Updated it there to prevent this in the future.)

Issue #1624, opened 2022-02-09
Paste from other app does not work (Mac OS)


murakamiriyugi:
Pasting works fine when copied from wxMaxima window.
However, copying from other applications (e.g., from sublime)
results in pasting nothing. My clipboard tells me that copying
is successful but it cannot be pasted on WXMaxima window.
Only copied texts from wxMaxima window can be pasted.

OS: macOS Monterey 12.12 (M1)
wxMaxima 21.11.0
wxWidgets 3.1.5
Maxima 5.45.1
SBCL 2.1.9

wxMaxima was installed thru home-brew

gunterkoenigsmann:
Looked at the code. wxMaxima asks the clipboard if it can deliver text and when this is true pastes the text. My guess is that something in MacOs has changed in a way that is still unknown to wxWidgets' clipboard handling. Unfortunately I don't own a Mac I can test it on.

k-yaegashi:
Same issue here. I debugged wxMaxima and found that `wxTheClipboard->IsSupported(wxDF_TEXT)` always returns false at line [7471](https://github.com/wxMaxima-developers/wxmaxima/blob/main/src/Worksheet.cpp#L7471) and [7575](https://github.com/wxMaxima-developers/wxmaxima/blob/main/src/Worksheet.cpp#L7575) in [src/Worksheet.cpp](https://github.com/wxMaxima-developers/wxmaxima/blob/main/src/Worksheet.cpp) ([`Worksheet::PasteFromClipboard()`](https://github.com/wxMaxima-developers/wxmaxima/blob/main/src/Worksheet.cpp#L7461)). In order to inspect in detail, debugging wxWidgets built from source seems to be also required.

OS: macOS Big Sur 11.6 (x86_64)
wxMaxima 22.03.0
wxWidgets 3.1.5
Maxima 5.45.1
SBCL 2.1.9

gunterkoenigsmann:
Resolved with commit 2b41864d8b53138428c49332e45caf0f279e6dff

PR #1647, opened 2022-04-04
Fix bug #1624: Can paste text from other apps in macOS

k-yaegashi:
Fixes the bug #1624.

The type of clipboard texts from other Apps is usually `kUTTypeUTF8PlainText = "public.utf8-plain-text"` corresponding to `wxDF_UNICODETEXT`, while current implementation only accepts type of `kUTTypeTraditionalMacText = "com.apple.traditional-mac-plain-text"` corresponding to `wxDF_TEXT`. Therefore accepting `wxDF_UNICODETEXT` solves this problem.

### DIFF
--- a/src/Worksheet.cpp
+++ b/src/Worksheet.cpp
@@ -7468,7 +7468,10 @@ void Worksheet::PasteFromClipboard()
     return;
 
   // Check if the clipboard contains text.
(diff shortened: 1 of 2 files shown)


gunterkoenigsmann:
Thanks a lot!

PR #1675, opened 2022-05-20
A minor grammar fix

richardgobeli:
A minor grammar fix.  Makes it easier to understand.


gunterkoenigsmann:
Thanks a lot! I am afraid we'll need loads of fixes like this!

PR #1712, opened 2022-09-11
Make various dialogs window-modal

minoki:
On macOS, wxMaxima uses a single process to show all document windows.  A consequence is, use of `ShowModal` make background windows unusable.

This PR improves the situation by making various dialogs use `ShowWindowModalThenDo`, which allows other windows to interact with user.


gunterkoenigsmann:
Thanks a lot!
Did already do much works to get rid of as many modal dialogues as possible.

PR #1906, opened 2024-03-27
Better dependencies for our po4a call

gunterkoenigsmann:


gunterkoenigsmann:
Wolfgang had a better idea, here.

PR #1916, opened 2024-05-09
SystemWiz: fix invalid format specifier

asakovets:
Otherwise, an assertion fails, when trying to solve a linear system (Equations > Solve symbolically > Solve Linear System...):

> /usr/include/wx-3.2/wx/strvararg.h(484): assert "(argtype & (wxFormatStringSpecifier<T>::value)) == argtype" failed in wxArgNormalizer(): format specifier doesn't match argument type


gunterkoenigsmann:
Thanks a lot!

PR #2008, opened 2025-04-04
Custom dynamic memory limit

Murmele:
Add possibility to easily overwrite the maxima dynamic memory limit in the gui

grafik

Currently the label is not vertically aligned. How can I achive this?

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -801,6 +801,8 @@ void Configuration::ReadConfig() {
   config->Read("greekSidebar_Show_mu", &m_greekSidebar_Show_mu);
   config->Read("symbolPaneAdditionalChars", &m_symbolPaneAdditionalChars);
   config->Read("parameters", &m_maximaParameters);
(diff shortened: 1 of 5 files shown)


daute:
Hi,
thanks for the patch. This is only necessary/effective if the SBCL Lisp compiler is used. And it is not changed very often (I assume).
Would it be better to just add an example line (as `-l lispname`) in the gui? E.g.
`-X '--dynamic-space-size nnn'`   Set the SBCL memory limit to nnn Megabytes?

As we have already described in the manual?
Best regards, Wolfgang

Murmele:
> ne (as `-l lispname`) in the gui? E.g.

Good point, I will change it. Yes it will be set only once, but if you have multiple computers you have to set it everywhere and I have always to search where I find the command.

daute:
That was already coded, but had a non working check for SBCL.
I disabled this check (for now), it is documented, that these options are only valid for SBCL.

daute:
I added the possible SBCL options as text:
wxmaxima-options-suggestion
Is it okay, to reject your pull request?

Murmele:
This is fine for me. Thank you!

Issue #1917, opened 2024-05-13
build error on Manjaro: missing wxmaxima.zh_CN.md


IInatas:
hi all together

will updating in Manjaro it tried to update to version wxmaxima 24.05.0-1 (Mo 13 Mai 2024 07:57:41 CEST), but the process failed.
The system contains the following tools:
-- Found pandoc: /usr/bin/pandoc (found version "3.1.8")
-- Found LATEX: /usr/bin/latex  found components: XELATEX LUALATEX 
-- Found Gettext: /usr/bin/msgmerge (found version "0.22.4")
-- Found Maxima: /usr/bin/maxima (found version "5.47.0")
-- Found appstreamcli: /usr/bin/appstreamcli (found version "1.0.2")

the unpacking was smooth and the compile run is starting just fine, but after 9%
....
[ 10%] Embedding toleranceCalculations.wxm
[ 10%] Built target build_toleranceCalculations.h
[ 11%] Generating wxmaxima.zh_CN.html
pandoc: wxmaxima.zh_CN.md: withBinaryFile: does not exist (No such file or directory)
make[2]: *** [info/CMakeFiles/build_wxmaxima.zh_CN.html.dir/build.make:73: info/wxmaxima.zh_CN.html] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:1376: info/CMakeFiles/build_wxmaxima.zh_CN.html.dir/all] Fehler 2
make[1]: *** Waiting for unfinished processes …
[ 11%] Built target build_manualanchors.h
make: *** [Makefile:166: all] Error 2

So far it looks like the file wxmaxima.zh_CN.md is missing. The file wxmaxima.zh_CN.html is present, so may be the is no need to build it from the md file?

I would like to receive instructions on how to proceed in order to perform a successful build run.  

regards,
Marcus

daute:
Please install po4a too. (https://packages.manjaro.org/?query=po4a). The language specific Markdown files are generated using po4a.
Best regards, Wolfgang

daute:
Hm. I tried to 'fix' the issue when po4a is not installed now, but it works properly.
If it is installed you should get a message like
```
-- Found po4a: /usr/bin/po4a (found version "0.57")
```
(0.57 is rather old, yes), or:
```
-- po4a not found. Disable the generation of localized documentation.
```

What do you get? Here the build works in both cases. Can you post the complete output of the cmake configuration run for wxMaxima (when using an empty build directory?)
[...]

PR #2067, opened 2026-02-10
cmake: alternative-fix for file_RELATIVRPATH

jgmbenoit:
This patch actually deals with upstream issue [#1917](https://github.com/wxMaxima-developers/wxmaxima/issues/1917). However the implemented fix fails in the Sid environment. This patch suggests an alternative fix that does its job in Sid.


daute:
Hmm. The patch would produce `file(RELATIVE_PATH RELPATH_TO_PO4A_PO_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_SOURCE_DIR}/locales/manual")` - which would be a constant. What are the issues in Debian sid?
I checked https://tracker.debian.org/pkg/wxmaxima but could not find a (broken) build log.

I am including this pull request for now. Let's see, if our CI scripts work...

(And what's the issue reported with reproducible builds there? We use UTC when generating timestamps. (`file(TIMESTAMP ${CMAKE_CURRENT_SOURCE_DIR}/wxmaxima.1.in MANPAGEDATE "%Y-%m-%d" UTC)`)

PR #2122, opened 2026-07-22
EditorCell split slice 5+7: move Maxima-code-only state onto CodeEditorCell

gunterkoenigsmann:
## What

Completes the long-running **EditorCell split** arc (slices 1–4, 6 already landed). The tokenizer output, autocomplete word list and matching-parenthesis highlight are only meaningful for Maxima **input** cells, yet every `EditorCell` — text and every heading level included — carried that state. This moves the state *and its behaviour* onto the `CodeEditorCell` subclass, so prose/heading cells no longer carry it.

**Behaviour is unchanged** — this is a pure relocation behind virtual dispatch.

## The move

The code-only cluster is inseparable (`FindMatchingParens`/`FindMatchingQuotes` iterate the token list; `StyleTextCode` populates both the tokens and the word list), so it moves as a unit onto `CodeEditorCell`:

- **State:** `m_tokens`, `m_tokens_including_hidden`, `m_wordList`, `m_paren1`/`m_paren2`, and their validity flags.
- **Methods:** `StyleTextCode()`, `FindMatchingParens()`, `FindMatchingQuotes()`, `GetAllTokens()`, and the paren-highlight drawing.

Shared editing/drawing code in the base reaches the moved behaviour through **five virtual seams**, each a no-op / empty default on prose:

| Seam | Called from |
|---|---|
| `FindMatchingParens()` | shared cursor/edit operations |
| `GetAllTokens()` | tokeniser consumers (empty for prose) |
| `GetWordList()` | autocomplete gathering (empty for prose) |
| `ClearParenMatch()` | the shared paren-reset sites |
| `DrawParenHighlight()` | `Draw()` |

Because `StyleTextCode()` moved onto the subclass, the base layout helpers it uses (`m_text`, `GetTextSize`, `HandleSoftLineBreaks_Code`, `BreakAfterOperator`, the `SoftBreakCandidate` struct) are now `protected`.

## Slice 7 (dead-branch cleanup)

`FindMatchingParens()` is now a no-op on prose, so the eight `if (GetType() == MC_TYPE_INPUT) FindMatchingParens();` guards became dead. All ten call sites now call it unconditionally, matching the two that already did. (The build compiling proves no base branch touched the moved state; the remaining `m_type == MC_TYPE_INPUT` conditionals are live behaviour and are intentionally left alone.)

## Verification

Builds clean (first try), and the EditorCell unit tests pass:

- `test_EditorCellInvariants` — **96,435 assertions** (the fuzz driver: thousands of real key events exercising paren-matching and styling across **both** code and text cells)
- `test_GroupCellLayout` — 55 assertions (covers the code-vs-prose virtual dispatch)
- `test_TreeUndo`, `test_WXMRoundtrip`, `test_WXMXRoundtrip` — all green

CI will run the full suite + the ASan/UBSan sanitizer build.

https://claude.ai/code/session_01DU23YMfQHdrkGCNJxn7dJs

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -158,6 +158,7 @@ stability fixes.
 - Windows: wxMaxima now repairs its own .wxmx/.wxm/.mac file association on startup, so updating to a new install location no longer makes Windows "forget" which program opens .wxmx files.
 - Windows: a new Help menu item registers wxMaxima as the tool TortoiseSVN and TortoiseGit use to compare .wxmx files.
 - Added a regression test guarding the worksheet's line-wrapping (line-breaking) layout pass.
+- Internal: the Maxima-code-only editor state (tokenizer output, autocomplete word list and matching-parenthesis highlight) now lives on the CodeEditorCell subclass instead of on every EditorCell, so text and heading cells no longer carry it. Behaviour is unchanged.
 
 # 26.07.0
 
--- a/src/cells/EditorCell.cpp
+++ b/src/cells/EditorCell.cpp
@@ -127,6 +127,33 @@ public:
   //! Code serializes to RTF snippet-by-snippet, keeping the syntax
   //! highlighting; prose maps its heading/text style in the base.
   wxString ToRTF() const override;
+
+  // --- code-only state + behavior (moved off the base in slice 5) ---
+  const std::vector<wxString> &GetWordList() const override { return m_wordList; }
+  const MaximaTokenizer::TokenList &GetAllTokens() const override;
+  void FindMatchingParens() override;
+  void ClearParenMatch() override { m_paren1 = m_paren2 = -1; }
+  void DrawParenHighlight(wxDC *dc) override;
+
+private:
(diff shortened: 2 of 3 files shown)


gunterkoenigsmann:
Claude said:

Honest note on value: as I flagged before you greenlit it, this is a lateral-encapsulation change — it trades ~88 bytes of unused state per text cell for a slightly leakier protected base surface (five helpers promoted to protected). It's clean and correct, but it's a data-model tidy-up, not a stability or perf win. That closes the EditorCell decomposition arc; I'd not recommend chasing the remaining live m_type==INPUT conditionals — each would be its own convert-to-virtual slice with the same lateral payoff.

=> Unsure if we really want to merge this one or if I was too enthusiastic about the slightly cleaner data model.

gunterkoenigsmann:
Perhaps I should have listened to Claude's warning and not pursued that plan: in the end it didn't pay off.

PR #2131, opened 2026-07-25
HTML export: use the MathJax 3 config API so equation settings apply again

gunterkoenigsmann:
Fixes the config/version mismatch spotted while planning the self-contained HTML export.

## The bug
The HTML export configures MathJax with **`MathJax.Hub.Config({...})`** — the **MathJax 2** API — but the loader URL it emits pulls in **MathJax 3** (`cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js`). MathJax 3 **ignores** `MathJax.Hub.Config` entirely — it reads `window.MathJax`, assigned *before* the loader runs. So the export's **left-aligned equations** and **left-side equation tags** have silently done nothing (they've been rendering with MathJax 3's defaults: centered, right-side tags).

It's a textbook symptom of "config written for an old version, never updated when the version moved."

## The fix
Emit the MathJax 3 configuration instead — assign `window.MathJax` before the loader script:

```html
<script>
  window.MathJax = {
    tex: {tagSide: "left"},
    chtml: {displayAlign: "left"},
    svg: {displayAlign: "left"}
  };
</script>
<script id="MathJax-script" async src="…mathjax@3/es5/tex-mml-chtml.js"></script>
```

- `TeX: {TagSide: "left"}` → `tex: {tagSide: "left"}`
- `displayAlign: "left"` → set under **both** `chtml` (the CDN default output) and `svg`, since `displayAlign` is an output-renderer option and a user can point `MathJaXURL` at an SVG build.
- Dropped `context: "MathJax"` (a MathJax-2-only menu option).

## Verification
`WorksheetExport.cpp` compiles clean. The visible effect (equations left-aligned again in exported HTML) is best confirmed by exporting a worksheet in a MathJax equation mode and opening it — happy to do that under the run-wxmaxima harness if you want, but it's a mechanical config-string fix.

Context: this is item #1 from the "keeping MathJax current" discussion; the standing "nudge us on new MathJax majors" CI job and the native-MathML self-contained export are separate follow-ups.

https://claude.ai/code/session_01DU23YMfQHdrkGCNJxn7dJs

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@
   instead of the generic wxWidgets "W" icon. The window icon now also falls back
   to the logo compiled into the binary, so it no longer depends solely on the
   .ico resource being the one Windows selects.
+- HTML export: the MathJax equation configuration (left-aligned equations and
+  left-side equation tags) is applied again. The export still used MathJax 2's
+  MathJax.Hub.Config() call, which the MathJax 3 that is actually loaded ignores,
+  so those settings had silently stopped taking effect; it now uses the MathJax 3
+  configuration.
 
 # 26.07.1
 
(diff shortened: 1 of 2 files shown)


gunterkoenigsmann:
Will be superseded by switching to MathJaX4

PR #2182, opened 2026-08-03
Add real tab-character support to EditorCell

gunterkoenigsmann:
## Summary

Previously `EditorCell::TabExpand()` immediately rewrote every `'\t'` into 1-4 literal spaces the moment it reached an `EditorCell` (typing, pasting, or loading a file), so a tab never actually existed in `m_text` — and a plain `.mac` file's tabs were silently lost on open/save. A tab is now a real character in `m_text`, expanded to the next 4-column tab stop only where text becomes pixels or a click becomes a text offset:

- `MaximaTokenizer` now always emits a tab as its own isolated, single-character token (like a newline), never merged into a space run — the load-bearing simplification that lets every `m_styledText`-based site (`Draw()`, `Recalculate()`, `GetLineWidth()`, `SelectPointText()`'s code-cell branch, `StyleTextCode()`) just compare `text == "\t"` instead of splitting substrings.
- `StyleTextTexts()` (the prose/text-cell path, which doesn't tokenize) gets the same isolation via a new `PushTextLine()` helper.
- Two new helpers, `EditorCell::NextTabStop()` and `MeasureTextWidth()`, do the actual tab-stop math. Sites that measure a raw `m_text` substring rather than a single token (`MarkSelection()`, `MixedDirectionOffset()`, `StyleTextTexts()`'s wrap check) go through `MeasureTextWidth()`, which splits on `'\t'` internally.
- Tab width is position-dependent — the one thing `GetTextExtent()` can't compute on its own — so a tab token's width is never cached the way `StyledText::SetWidth()` caches every other token.
- The Tab key (no selection) now inserts a real tab instead of spaces; selecting one or more lines and pressing Tab/Shift+Tab indents/dedents with a real tab per level, falling back to removing up to 4 leading spaces on already space-indented text.
- Left/Right arrow and Delete needed no changes: they already move/delete exactly one `m_text` character, which is now correct automatically.
- Removed `WXK_BACK`'s old "gobble up to 4 trailing spaces" shim, a workaround for the space-expanded-tab world that would otherwise misfire on real spaces the user actually typed.
- `TabExpand()` is replaced by `NormalizeLineEndings()`, keeping only its `"\r\n"` → `"\n"` normalization (a lone `'\r'` is wxMaxima's own soft-break marker and must stay untouched).

Bidi needed no changes: `'\t'` is Unicode bidi class Segment Separator, already resolved as a neutral character by `fribidi_log2vis()` with no special-casing — confirmed by the existing `EditorCellBidi` tests still passing unmodified.

## Test plan

- [x] `test_EditorCellTabs.cpp` (new): a tab survives `SetValue()`/`GetValue()` untouched, `GetLineWidth()` tab-stop alignment at various starting columns and for consecutive tabs, Tab-key insertion, Backspace removing a whole tab in one press (and plain spaces still deleting one at a time), selection indent/dedent with a real tab (and dedent falling back to space-removal on already space-indented text).
- [x] `test_MACRoundtrip.cpp` (new): a tab used for interior alignment in a code cell survives `Format::TreeToWXM(wxm=false)`/`ParseMACContents()` byte-for-byte — the same code `Worksheet::ExportToMAC()`/`MaximaFileIO::OpenMACFile()` use for plain `.mac` files.
- [x] Existing `EditorCellWrapping`, `EditorCellBidi`, `EditorCellInvariants`, the `Worksheet*` suite, `WXMRoundtrip`, and `WXMXRoundtrip` all still pass unmodified.
- [x] Full local build clean (zero warnings) with `g++-11 -Wall -Wextra` (matching CI's `compile_2204`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01UDdgx8DR7G9w9B5hBZAo9R)_


gunterkoenigsmann:
Pushed a fix (0b8b995) for a bug found while testing this: Tab/Shift+Tab on a multi-line selection collapsed the selection immediately afterward (`CursorPosition(start)` was called right after `SetSelection(start, end)`, and `CursorPosition(pos)` sets both selection ends to `pos`), so a second Tab/Shift+Tab press without re-selecting silently did nothing instead of indenting/dedenting further. This predates the tab-character work in this PR (identical in the parent commit) but sits in code this PR actively touches, so fixed it here rather than filing separately. Added a regression test (repeated Tab/Shift+Tab without re-selecting).

Also looked into the "Compile using minGW" failure notification on this PR: its `HeadSHA` (`d04a5e9`) is an older commit from the already-merged #2181, not this PR's current head — a stale/late-arriving result, not something to act on here.

---

gunterkoenigsmann:
Investigated the `compile_and_test_sanitized` failure (for commit `74381b8`, since superseded by later pushes — fresh CI is already running against the current head). 3 tests failed there: `lisp_mode` (Failed), `openMacFiles` (Timeout), `tutorial_10Minutes` (Timeout). None are related to this PR:

- `lisp_mode`: known pre-existing flake on `main`, unrelated to any particular change (confirmed in earlier PR investigation this session).
- `openMacFiles`: reproduces identically (same timeout) on an unmodified `main` checkout in an isolated worktree, confirmed directly in a local sandbox rebuild — pre-existing.
- `tutorial_10Minutes`: same failure shape (batch-mode Maxima evaluation via `--exit-on-error` timing out at 380s) as `openMacFiles`, alongside it in the same run — almost certainly the same class of pre-existing CI-runner flakiness, not something touched by `EditorCell`/`MaximaTokenizer` changes.

Meanwhile every test that actually exercises this PR's changes passed in that same run: `EditorCellTabs`, `MACRoundtrip`, `WXMRoundtrip`, `WXMXRoundtrip`, `EditorCellWrapping`, `EditorCellBidi`, `EditorCellInvariants`.

---

gunterkoenigsmann:
Root-caused the CI redness on this PR (`compile_and_test_sanitized`, `compile_latest_and_test`, `Compile using minGW` — all failing/timing out on `lisp_mode`/`openMacFiles`/`tutorial_10Minutes`/`rememberingAnswers`/`testbench_simple.wxmx`) and filed it as #2183: `--batch --exit-on-error` hangs forever when a worksheet provokes an interactive Maxima question, instead of halting the way its own `--help` text promises. Confirmed with a live `gdb` trace (not a guess) — one cell in `test/testbench_simple.wxm` triggers a genuine Maxima question mid-evaluation (`"Enter space-separated numbers, 'all' or 'none': "`), and wxMaxima has no logic to treat that as the documented halt condition, so it just waits forever. Full diagnosis, repro, and evidence are in #2183.

This is unrelated to this PR's diff (reproduces identically on unmodified `main`) and confirmed reproducing across every commit pushed here, on both Linux and Windows CI. Nothing further needed from this PR for it — tracking continues in #2183.

---

PR #2219, opened 2026-08-06
Fix wx-cd failing on wildcard, bracket, and non-ASCII directory names (GH #1672)

### LINKED ISSUE #1672: Working directory message
Hi,

I encountered the following red message when I was preparing a document in wxMaxima:

`Warning: Can set maxima's working directory but cannot change it during the maxima session`

It appears in different locations in the document and I can't get rid of it. What can I do to remove it ?

gunterkoenigsmann:
Fixes #1672.

`wx-cd` (the Lisp function wxMaxima uses to keep Maxima's working directory in sync with the worksheet's, in `src/wxMathML.lisp`) had two independent bugs and a nuisance that compounded each other into the reported "hundreds of copies of the same warning" symptom:

1. **Wildcard/bracket characters.** `pathname-directory` parses a directory string using Common Lisp's *portable* pathname syntax, which treats `*`, `?`, `[` and `]` as wildcard markers rather than literal characters. A directory legitimately named e.g. `Draft [v2]` turned into an actual wildcard pattern, and `chdir`-ing to a wildcard pattern always fails. Fixed (SBCL) by parsing with `sb-ext:native-pathname` instead, which round-trips those characters literally (verified directly with `sbcl --script` against real bracket/wildcard/Unicode directories).

2. **Windows codepage mismatch.** `sb-posix:chdir`/`sb-posix:getcwd` go through the C runtime's ANSI-only `_chdir`/`_getcwd`, which encode/decode the path using the OS's *legacy codepage* rather than Unicode. A directory name containing a character outside that codepage (Cyrillic, CJK, ...) fails even though it genuinely exists — this is a confirmed, still-open upstream SBCL bug ([launchpad #2100706](https://bugs.launchpad.net/sbcl/+bug/2100706)). Fixed by calling the Win32 wide-character APIs (`SetCurrentDirectoryW` / `GetCurrentDirectory`) directly via SBCL's alien FFI, with the previous `sb-posix`-based behavior kept as an automatic fallback if the wide-API call is unavailable or errors. This code is Windows-only and couldn't be exercised on the Linux sandbox it was developed in, so it's designed defensively: a mistake here can only make the bug persist, never stop Maxima from starting.

3. **Repeated warnings.** Once a path was persistently broken, `wx-cd` re-printed the identical warning on every failed cd attempt for the rest of the session. Since a saved `.wxmx` keeps every cell's output verbatim, reopening and resaving an affected document across many sessions accumulates one new copy per session — matching what's reported in the issue. `wx-cd` now remembers the last directory it failed on and skips a repeat warning for the same failure, resetting once a cd succeeds or a different directory fails.

## Testing

- Verified the wildcard/bracket/Unicode fix directly against real directories using `sbcl --script` (bracket, wildcard, semicolon, and CJK/Cyrillic/accented-Latin directory names all now `chdir` correctly; confirmed they failed beforehand).
- Verified the "warn once per distinct failure" dedup logic the same way (repeat failure on the same path is silent; a different path or a recovery re-arms it).
- Added `test/check-wx-cd-non-ascii.cmake` + a new `check-wxMathML.lisp_wx-cd_non_ascii_dir` ctest that `chdir`s into a directory with a CJK name and asserts it succeeds. It's a CMake script rather than the existing `check-wxMathML.sh` POSIX-shell helper specifically so it isn't excluded by the `needs_posix` label and actually runs on the Windows CI job too. Ran it (and the full `check-wxMathML.*` group, to check for regressions) locally — all pass.
- Full local build succeeds; no C++ was touched.
- The Windows-specific `SetCurrentDirectoryW` codepath itself (item 2) could not be exercised locally (no Windows environment available) — it's additive/fallback-guarded for that reason.

🤖 Generated with [Claude Code](https://claude.ai/code)

---

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,31 @@
 # Current development version
 
+- "Can set maxima's working directory but cannot change it during the
+  maxima session" (GH #1672): this warning, printed by `wx-cd` whenever it
+  fails to keep Maxima's working directory in sync with the worksheet's,
+  had two separate root causes and a nuisance of its own, all fixed:
+  - A directory name containing `*`, `?`, `[` or `]` (e.g. `Draft [v2]`)
+    always failed: Common Lisp's portable pathname syntax treats those
+    characters as wildcard markers rather than literal ones, so SBCL's
+    `pathname-directory` turned a perfectly real directory into an
+    unmatched wildcard pattern. Fixed by parsing the string with
+    `sb-ext:native-pathname` instead, which has no such reinterpretation.
+  - On Windows specifically, a directory name containing a character
+    outside the system's active codepage (e.g. Cyrillic or CJK on a
+    Western-European Windows install) also always failed:
+    `sb-posix:chdir`/`sb-posix:getcwd` go through the C runtime's
+    ANSI-only `_chdir`/`_getcwd`, which cannot represent such characters
+    at all (a confirmed, still-open upstream SBCL bug, launchpad
+    #2100706). Fixed by calling the Win32 wide-character APIs directly
+    instead, with the previous behavior kept as an automatic fallback.
+  - Once triggered by a persistently broken path, the warning printed
+    again on every single cd attempt for the rest of the session, and
+    since a saved `.wxmx` document keeps every cell's output verbatim,
+    reopening and resaving an affected document over many sessions could
+    accumulate hundreds of copies of the exact same warning. `wx-cd` now
+    remembers the last directory it failed on and only reports a given
+    failure once, until either a different directory fails or the same
+    one starts working again.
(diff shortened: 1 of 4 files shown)


gunterkoenigsmann:
`setup_cygwin` failing here isn't from this PR's changes: it's failing the same way intermittently on `main` right now too (e.g. runs [31119371305](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/31119371305) and [31116184247](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/31116184247), both today), and in both cases the failure is in GitHub's own "Set up job" step before any of this repo's code runs — runner/infra flakiness, not a real build failure. Will re-run once the base branch's `compile_cygwin` job recovers.

---

PR #2222, opened 2026-08-06
Fix translation files not reaching the macOS app bundle (GH #1711)

### LINKED ISSUE #1711: macOS: Translation files are not copied into the application bundle
On macOS, translations are not applied because `*.mo` files are not in the application bundle.

One would need to copy `${CMAKE_BINARY_DIR}/locale` (generated by `update-locale`) to `${CMAKE_BINARY_DIR}/src/wxmaxima.app/Contents/Resources/locale`.

gunterkoenigsmann:
Fixes #1711.

**I don't have a Mac to test this on, so please treat this as needing verification on the next macOS build/DMG before considering it closed.**

Two independent bugs meant `.mo` files never reached `Contents/Resources/locale/<lang>/LC_MESSAGES/`, where `Dirstructure::LocaleDir()`/`wxFileTranslationsLoader` (`main.cpp`) actually look for them at runtime:

1. **`locales/wxMaxima/CMakeLists.txt`'s `copy_mo_file_${LANG}_for_wxmaxima_local` target** — the only thing that populates `${CMAKE_BINARY_DIR}/share/locale/<lang>/LC_MESSAGES/wxMaxima.mo` with the correct per-language layout, and `ALL`-tagged so it runs on every normal build — was unconditionally skipped on Apple, working around what its own comment says is an Xcode-generator issue. The macOS CI job that actually produces the shipped DMG uses `-G Ninja`, not Xcode (only a separate, non-packaging smoke-test job uses Xcode), so I narrowed the guard to just the Xcode generator instead of excluding all of Apple. Confirmed live (on Linux, but the CMake logic itself isn't platform-specific): a plain `ninja`, having never explicitly run `update-locale`, already produces a fully populated `build/share/locale/*/LC_MESSAGES/*.mo` from this target alone — so on macOS it was producing nothing at all.

2. **Separately, `src/CMakeLists.txt`'s macOS bundle resource list** tried to `file(GLOB ${CMAKE_BINARY_DIR}/locale/*.mo)` into the `RESOURCE` target property. This can't work for two stacked reasons: `file(GLOB)` freezes its result at *configure* time, before a single build step has run and generated any `.mo` file (translations are generated during the build); and the pattern is non-recursive, so it wouldn't have matched the actual nested `locale/<lang>/LC_MESSAGES/wxMaxima.mo` layout even if the files existed by then. And even if both of those were fixed, CMake's `RESOURCE` property flattens everything it's given directly into `Contents/Resources`, with no way to reproduce a subdirectory structure. Fixed by dropping the glob and instead copying the (now-populated, already-correctly-nested) `share/locale` directory into `Contents/Resources/locale` via a plain `file(COPY ...)` inside the existing `install(CODE ...)` block that already runs `fixup_bundle` — the same "has to be a separate step, everything the build produces is only guaranteed to exist by then" reasoning that block's own comment already gives for deferring `fixup_bundle` itself.

## Testing

- CMake configure and a full build succeed unaffected on Linux (the guard change is a no-op there: `CMAKE_GENERATOR` is never `"Xcode"` outside Apple), and `share/locale` still populates correctly.
- Validated the `file(COPY ...)` logic standalone via `cmake -P` against a fake `share/locale/<lang>/LC_MESSAGES/*.mo` tree, confirming it reproduces `Contents/Resources/locale/<lang>/LC_MESSAGES/wxMaxima.mo` correctly.
- The Apple-only code paths themselves (the `if(APPLE)` branches in both files) are **not** exercised by any of the above and remain unverified on real hardware. The `file(COPY ...)` step is guarded by `if(EXISTS ...)`, so if something about this is still wrong on macOS specifically, it should fail silently (no translations, same as today) rather than break the build.

🤖 Generated with [Claude Code](https://claude.ai/code)

---


gunterkoenigsmann:
The "Compile using minGW" failure isn't from this PR's own diff: `check-wxMathML.lisp_box_no_extra_parens`, `_highlight_box_keeps_parens`, and the two `_sum_*_summand_*` tests all use the POSIX-shell `check-wxMathML.sh` helper but were never added to `WXM_POSIX_TESTS` (the allowlist that tells Windows CI which tests to skip rather than try to run as a `.sh` file directly) — a pre-existing gap on `main` since the box-parens PR, just never caught by a Windows CI run until now. Pushed a fix (added the missing names, plus a comment pointing future additions at the allowlist) in 674669f.

---

Issue #2286, opened 2026-08-31
wxAppIndicator


gunterkoenigsmann:
We don't seem to use that on gtk, currently.

Issue #2282, opened 2026-08-20
Worksheet displays empty braces { } for odelin() result, but content exists (visible when copying as LaTeX with broken \left{ / \right} escaping)


michelprog59:
**wxMaxima version: 26.07-1
**Maxima version: 5.49
**OS: Linux Ubuntu 24.04

<img width="480" height="270" alt="Image" src="https://github.com/user-attachments/assets/6a06f955-6b06-449d-b7ee-76b72173d113" />

**Description**
When solving an ODE with odelin() from the contrib_ode package, the
worksheet displays an empty set of braces `{ }` for the result —
no mathematical content is rendered inside them. However, the
content is NOT actually empty: copying the cell as LaTeX reveals
that the expression exists but is exported with invalid, unescaped
braces (\left{ ... \right} instead of \left\{ ... \right\}).

This suggests two related but distinct issues:
1. A rendering/display bug: the on-screen worksheet cell fails to
   draw the content inside the braces, even though the underlying
   expression is not empty.
2. A LaTeX export bug: when copying this cell as LaTeX, the braces
   \left and \right are not escaped, producing invalid LaTeX syntax.

**Steps to reproduce**
1. load(contrib_ode);
2. eq3: (x^2 + 1)*'diff(y,x,2) + x*'diff(y,x) - y = 0;
3. sol3: odelin(eq3, y, x);
4. Observe the worksheet: the result is displayed as empty braces
   { } with no visible content between them (see attached screenshot).
5. Select the cell and copy it as LaTeX (Edit > Copy as LaTeX, or
   right-click > Copy as LaTeX).
6. Paste the copied code into a .tex file and try to compile it.

**Actual result**
- On screen: empty braces { } are displayed, with the actual solution
  content invisible/not rendered (screenshot attached).
- In the copied LaTeX: the hidden content becomes visible as text,
  but is malformed:

  \left{ x\mathop{,}\sqrt{{{x}^{2}}\mathop{+}1}\right}

  This does not compile because \left and \right require escaped
  braces (\left\{ ... \right\}).

**Expected result**
- On screen: the worksheet should display the actual set/expression
  content inside the braces, not an empty pair of braces.
- In LaTeX export: braces should be properly escaped:

  \left\{ x,\sqrt{{x}^{2}+1}\right\}

**Additional notes**
This looks like it could be a single underlying cause: something
in the cell layout/rendering code is not correctly measuring or
drawing the size of the set expression's content when a { } set is
[...]

Issue #2270, opened 2026-08-13
Datatype "set" is broken


ReneSci:
If I try to create a "set" it dosen't show anything:

<img width="670" height="262" alt="Image" src="https://github.com/user-attachments/assets/33aeee9a-f784-4ea3-9b89-43654f4b63a0" />

And I can not work with the set created afterwards.

But if I copy the output cell, the set has been created:
{[1,1],[2,4],[3,9],[4,16],[5,25],[6,36],[7,49],[8,64],[9,81],[10,100]}

and listify(%) can reveal the set:

<img width="250" height="107" alt="Image" src="https://github.com/user-attachments/assets/be63ddad-a82f-4b42-8f00-ef4372c064e2" />

```
I found the error because I use a set to set xtics in a plot - it dosen't work anymore.
```

```
(%i11)	wxbuild_info()$
WxMaxima version:             26.08.0_MSW

Using wxWidgets version:      wxWidgets 3.3.3

Maxima version:               branch_5_49_base_882_ga91ebf9c9

Maxima build date:            2026-08-13 00:02:52

Host type:                    x86_64-w64-mingw32

System type:                  Win32 10.0.26200 X86-64

Lisp implementation type:     SBCL

Lisp implementation version:  2.6.7

WxMaxima help dir:            C:\maxima-current\share/doc/wxmaxima

----------

Maxima's idea of the directory layout is:

maxima-prefix:           C:/maxima-current

maxima-topdir:           C:/maxima-current/share/maxima/branch_5_49_base_882_ga91ebf9c9

maxima-imagesdir:        C:/maxima-current/lib/maxima/branch_5_49_base_882_ga91ebf9c9/binary-sbcl

maxima-sharedir:         C:/maxima-current/share/maxima/branch_5_49_base_882_ga91ebf9c9/share

maxima-srcdir:           C:/maxima-current/share/maxima/branch_5_49_base_882_ga91ebf9c9/src

maxima-demodir:          C:/maxima-current/share/maxima/branch_5_49_base_882_ga91ebf9c9/demo

maxima-testsdir:         C:/maxima-current/share/maxima/branch_5_49_base_882_ga91ebf9c9/tests

maxima-docdir:           C:/maxima-current/share/maxima/branch_5_49_base_882_ga91ebf9c9/doc

maxima-infodir:          C:/maxima-current/share/info

[...]

Issue #1907, opened 2024-04-01
wxm file format - no escaping of "*/"


daute:
While writing documentation about the file format (commit 7e411a044ae104ac915410af6cc826632a5a10e7) I saw, that WxMaxima does **not** escape the comment end sequence (`*/`) in wxm files.
So while the following wxm file (I had to zip it for the Github upload) can be read in in wxMaxima, one can not read it with `load()` or `batch()` in Maxima (that should be possible since Maxima 5.38):

[a.zip](https://github.com/wxMaxima-developers/wxmaxima/files/14821822/a.zip)

A title cell (most likely all cell types) with "`abc */`" is written as:

```
/* [wxMaxima: title   start ]
abc */
   [wxMaxima: title   end   ] */
```

One can also 'trick' wxMaxima to create additional cells by creating a title cell with the following input (all in **one** title cell):
```
 abc
   [wxMaxima: title   end   ] */

/* [wxMaxima: title   start ]
def
```

daute:
The same problem exists - of course - for the comment beginning sequence: `/*`
Create worksheet with a title cell containing `abc /*`, save it as a.wxm and Maxima will produce an error:
```
(%i1) batch("a.wxm");

read and interpret <path-to-directory>/a.wxm
Maxima encountered a Lisp error:

 SIMPLE-ERROR: parser: end of file in comment.

Automatically continuing.
To enable the Lisp debugger set *debugger-hook* to nil.
```

gunterkoenigsmann:
Gaah! The wxm format isn't _meant_ to be extended...

daute:
Well, but with the current data format one can not reliably save documents - see the example above, where one can "create" cells, with just a special formatted text/title cell. 
And concerning the 'Maxima can import it'-feature, one can of course not only create error messages but also mathematical effects.

`x:1;`
[Text/Title cell with the following contents: `abc */ x:2$ /* def `]
`x;`

What will be the value of x after Maxima imported that wxm-file?

The easiest thing would be to just save "`/*`" as "`\/\*`" and "`*/`" as "`\*\/`" in non-math-cells and change that back, after the document is read back in. Or maybe use HTML like hex codes, e.g. `&#42;` for the asterisk  and `&#47;` for the slash. Or some simliar looking Unicode characters. Don't know, what is the best solution.

If one already uses comment signs in title/text/... cellls and uses the wxm format, he/she already has a problem...

gunterkoenigsmann:
[...]

PR #2287, opened 2026-09-02
GH #2286 tray icon + GH #1907 .wxm injection fix + GH #2282 regression test

gunterkoenigsmann:
## Summary

This PR now contains four independent, unrelated commits pushed to the same branch:

### `wxmaxima_version_string` CI fix attempt (unverified on real Windows)

`wxmaxima --version`'s output wasn't reliably reaching ctest's pipe on the minGW Windows runner (exit code 0, but no matching stdout content) — failing on essentially every push since 2026-08-15. The stdout/stderr redirection code (needed because wxmaxima is a `WIN32`-subsystem binary with no stdio wired up by default) opened a throwaway `FILE` via `_fdopen()` and shallow-copied its struct onto the real `stdout`/`stderr` object, which can leave CRT-internal-only bookkeeping (buffering state, the per-stream lock) inconsistent with an object at a different address. Switched to `_dup2()`, the documented way to repoint an existing stream's descriptor without fabricating a second `FILE` object.

**Genuinely unverified** — no Windows environment available in this sandbox to test on; only confirmed the change cross-compiles cleanly with `x86_64-w64-mingw32-g++` under the same `-Wall -Wextra` flags `compile_windows.yml` uses for this job. The next CI run on this PR is the real test. Full reasoning in `AGENTS.md` if this doesn't help and someone picks it up later.

### GH #2286 — System tray icon

Adds a system tray/notification-area icon (`src/TrayIcon.{h,cpp}`) mirroring wxMaxima's busy status — the same information the status bar's own icon and, on Windows, the taskbar button's progress overlay already show — plus a small quick-access menu (Interrupt, Show wxMaxima, Exit). Uses the portable `wxTaskBarIcon`; on GTK/Linux it only renders as a genuinely visible icon when the linked wxWidgets was itself built with AppIndicator/Ayatana support (a property of the wxWidgets package, not something wxMaxima's own code needs to special-case — it transparently falls back to the older GtkStatusIcon/XEmbed mechanism otherwise).

Reuses `StatusBar`'s own already-decoded icons (new `StatusBar::GetTrayIconBitmap()` getter) rather than re-embedding the same art a second time — the generated `art/statusbar/*.h` headers declare their byte arrays without `static`/`extern`, so `#include`-ing one from a second `.cpp` file is a duplicate-symbol link error, caught on the first real link attempt. Menu actions (Interrupt/Exit) re-post plain `wxEVT_MENU` events to the main frame instead of duplicating the existing handlers, since `wxTaskBarIcon`'s own popup menu delivers events to itself, not the frame.

Verified live end-to-end in Xvfb with a real systray host (fluxbox with `systemtray` configured): the icon tracks real status transitions and matches the app's own status bar icon pixel-for-pixel; clicking "Interrupt"/"Exit" from the tray triggers the exact same real handlers a normal menu click would.

Only one new translatable string was needed (`"&Show wxMaxima"`) — the other menu items and two longer tooltips deliberately reuse existing wording verbatim (added as a single hand-edited `wxMaxima.pot` entry, not a full `update-locale` regen, which would have swept in ~1000 lines of unrelated pre-existing drift).

### GH #1907 — `.wxm` comment-injection security fix

A title/section/subsection/heading/comment cell's `.wxm` start marker opens a `/* ... */` comment that stays open across the cell's *entire* content, closing only at the end marker's own trailing `*/` — unlike an input/code cell's markers, which are each a single, already-closed one-line comment. A literal `*/` inside such a prose cell's own text therefore closed the comment early, turning everything up to the next `*/` in the file into live, executable Maxima input — silently, with no error — as soon as the file was opened or `batch()`/`load()`ed by plain Maxima.

Fixed with a reversible HTML-entity-style escape (`EscapeWXMSlashes()`/`UnescapeWXMSlashes()` in `src/WXMformat.cpp`): escape `&` first, then replace any `/` adjacent to a `*` with `&#47;` in a single linear scan, leaving ordinary slashes (e.g. `1/2`) untouched. Code/input cells are deliberately excluded, since they must stay byte-identical for a plain Maxima to parse. Also applied to the equivalent `.mac` (xmaxima interop) export path for defense in depth.

This closely follows an approach the maintainers had already discussed but never implemented in the issue's own comment thread.

Regression tests in `test/unit_tests/test_WXMRoundtrip.cpp` cover both the injection fix and the "code cells stay unescaped" invariant — confirmed to actually catch the bug by reverting just the source fix and re-running (the injection scenarios failed with exactly the reported symptom).

### GH #2282 — LaTeX brace-escaping regression test

While investigating #2282 I confirmed both parts of the report are already fixed on `main` (the display bug via #2270's fix, the LaTeX escaping via `924644cfe`), but the LaTeX fix had no test coverage. Added one `SCENARIO` in `test/unit_tests/test_LayoutInvariants.cpp` parsing a set containing a sqrt (the `odelin()`-shaped case from the bug report) through `MathParser` and asserting `SetCell::ToTeX()` emits `\left\{`/`\right\}`, not the invalid `\left{`/`\right}`. Confirmed it catches the regression by reverting `924644cfe` locally and re-running.

## Test plan

- [x] Full unit test suite (45 tests) passes, including `WXMRoundtrip`, `MACRoundtrip`, `WXMXRoundtrip`, `LayoutInvariants`, `GroupCellLayout`, `WorksheetLayout`.
- [x] `check-pot-coverage` passes with only the single new `"&Show wxMaxima"` string added.
- [x] Confirmed both #1907/#2282 test scenarios fail against pre-fix source and pass against the fixes.
- [x] TrayIcon verified live in a real Xvfb + fluxbox session (see above).
- [ ] `wxmaxima_version_string` fix — awaiting real Windows CI confirmation.

---


gunterkoenigsmann (commented):
**"Compile using minGW" CI failure is pre-existing on `main`, not caused by this PR.**

The job fails on a single test: `71 - wxmaxima_version_string (Failed)` -- `wxmaxima --version` doesn't produce output matching the expected `wxMaxima .*` regex on that runner. Neither of this PR's two commits touches `main.cpp`, the `--version` handling, or anything version/CMake-related, so this can't be something either commit introduced.

Checked the `compile_windows` workflow's history on `main` directly: it has failed on essentially every push since 2026-08-15 (`5dcee1e6`, "A speculative go at #2274") through the commit this PR is based on (`9e5b571`, 2026-09-01) -- ~18 consecutive red runs across many unrelated commits, with no fix landed yet. So this is confirmed pre-existing and unrelated to this PR, not a flake worth re-running (the failure is consistent, not intermittent).

For what it's worth, I looked at the actual code path: `main.cpp` already has fairly extensive Windows-specific plumbing to route `--version`'s `wxMessageOutput::Get()->Printf(...)` to stdout instead of a modal message box (`RedirectStdioToParent()`, `wxMessageOutput::Set(new wxMessageOutputStderr(stdout))`, both added specifically to fix this exact class of head-less-hang). Since the test still gets no matching output at all, something in that chain still isn't reaching stdout in this specific CI runner's environment -- but diagnosing that further needs a real Windows/minGW environment to reproduce in, which I don't have access to here. Not something I can fix confidently as part of this PR.

---

gunterkoenigsmann:
Status update on the `wxmaxima_version_string` CI failure this PR also attempted to fix (commit 3f8fd15, the `_dup2()` change): the next CI run reproduced the exact same failure, so that fix attempt is disconfirmed, not just unverified.

To check the theory properly instead of guessing again, I set up a Wine-based test harness in the sandbox (cross-compiled with `i686-w64-mingw32-g++`, run under Wine with a real `Xvfb`, stdout piped the same way ctest invokes the binary) and reproduced `BindStdStreamToParent()`'s exact old (struct-copy) vs. new (`_dup2()`) logic in isolation. Both variants deliver a GUI-subsystem process's piped stdout correctly in every combination tried (file redirect and anonymous pipe). So the bug isn't in that function's mechanism — something else in wxMaxima's actual Windows startup path is interfering with stdout before the `--version` branch's `Printf()` runs, in a way this minimal repro doesn't reproduce.

I kept the `_dup2()` change (it's still the more correct way to repoint a CRT stream regardless), but stopped describing it as a fix for this issue in NEWS.md/AGENTS.md. I documented the investigation and the most promising remaining lead (a timing correlation with the GH #2274 fix that reordered `OnInit()`) in AGENTS.md for whoever picks this up next — it likely needs either real Windows access or a full wxWidgets+wxMaxima MinGW build exercised under Wine, which was too large an undertaking for this pass.

This failure is pre-existing (not introduced by this PR) and unrelated to the tray icon/`.wxm` injection fix/regression test changes this PR actually makes, so I'm not blocking on it further here.


---

PR #2290, opened 2026-09-06
Add an opt-in MCP server exposing worksheet content as AI context

gunterkoenigsmann:
## Summary

Implements the request for "a de facto standard AI sidebar that allows to talk to an AI and gives it access to a worksheet" as an **MCP (Model Context Protocol) server** — scoped, per discussion, to a first, self-contained "server only" PR (an in-app chat sidebar is a separate follow-up decision).

**Safety scope, as agreed:** read-only worksheet context, with two explicitly approved exceptions. The server can never insert, edit, or evaluate anything in the worksheet.

- `list_cells` — every cell's UUID, type, a short input preview, and whether it has output
- `read_cell` — one cell's full input/output by UUID
- `read_worksheet` — the whole worksheet as text (size-capped)
- `read_toc` — the table of contents (title/section/.../heading6 cells, with level)
- `read_section` — one whole section's text by its heading's UUID (added after the suggestion that "AIs often additionally read the table of contents of big documents" — this is the TOC's natural companion, so an AI can navigate a large worksheet section by section instead of pulling everything at once)
- `read_variables` — the Variables sidebar's current watchlist (name/value pairs)
- `watch_variable` / `unwatch_variable` — add/remove a variable from that watchlist, confirmed as fine even though they're nominal writes: they only change what the sidebar tracks for display, exactly like a user typing a name into it by hand, and touch no worksheet content.

## Design

- **`McpTools`** (`src/mcp/McpTools.{h,cpp}`) is the actual tool logic, free of any networking/JSON-RPC — takes a `Worksheet*`/`Variablespane*`, returns `nlohmann::json`. Unit-tested directly (`test/unit_tests/test_McpTools.cpp`), no live Maxima or sockets needed.
- **`McpServer`** (`src/mcp/McpServer.{h,cpp}`) is the transport: MCP's "Streamable HTTP", the read-only subset that needs no server push — a single `POST /mcp` endpoint answering one JSON-RPC 2.0 request per connection (`GET` gets a plain 405). Built directly on `wxSocketServer`/`wxSocketBase`, event-driven on the GUI thread (the same primitive `Maxima.cpp` already uses for the Maxima protocol, just event-driven instead of threaded) — this lets it call straight into `Worksheet`/`GroupCell`/`Variablespane` with no cross-thread marshaling, since none of those are thread-safe.
- **Off by default**, one Options checkbox + a port field. Binds to `127.0.0.1` only, double-checks each connection's peer address, and validates the `Origin` header against localhost when a client sends one — the MCP spec's own called-out mitigation against a malicious web page DNS-rebinding an attacker-controlled hostname to `127.0.0.1`.
- JSON-RPC framing uses a vendored `nlohmann/json.hpp` (single header, MIT, v3.11.3, `src/vendor/nlohmann/`) rather than a hand-rolled parser, since real MCP clients need to interoperate with this correctly.
- `Variablespane` gained two small additive methods (`GetWatchedValues()`, `RemoveWatch()`) it didn't already expose publicly.

## Testing

- `test/unit_tests/test_McpTools.cpp`: headless coverage of every tool (cell listing/reading, TOC extraction, section boundaries including a nested-subsection case, the full watch/read/unwatch round trip) and its error paths (unknown tool/UUID, missing/invalid arguments) — 35 assertions, all passing.
- Live end-to-end verification in a real Xvfb session against a running, Maxima-connected wxMaxima: every tool exercised via `curl` (`initialize`, `tools/list`, `tools/call` for all eight tools), plus the transport edge cases (`GET`→405, bad `Origin`→403, good `Origin`→200, unknown path→404, a JSON-RPC notification→202), and a clean `SIGTERM` shutdown confirmed to leave no lingering process or open port.
- Full tree rebuild produces zero new warnings; the non-live-Maxima parts of the `ctest` suite pass (the live-Maxima-dependent batch tests fail in this sandbox for a pre-existing, already-documented environment reason unrelated to this change — see `AGENTS.md`).

Detailed design notes and the reasoning behind each decision are in `AGENTS.md`'s new "MCP server" entry.

---

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -272,6 +272,149 @@ working without extra checks.
 wxMaxima is a GUI front-end to the Maxima CAS; it talks to a Maxima process over
 a local TCP socket.
 
+- **MCP server (`src/mcp/McpServer.{h,cpp}`, `src/mcp/McpTools.{h,cpp}`,
+  2026-09) -- lets an external AI tool read the current worksheet as
+  context.** User request: "they say that there is a de facto standard on
+  how to create an AI sidebar that allows to talk to an AI and gives it
+  access to a worksheet" -- identified as MCP (Model Context Protocol).
+  Scoped down with the maintainer to a first, self-contained PR before any
+  in-app chat sidebar: an MCP server only, read-only except for two
+  explicitly-approved exceptions (see below), no write/insert/edit/evaluate
+  capability of any kind.
+  - **Split for testability**: `McpTools` is the actual worksheet-reading
+    logic (`ListCells`/`ReadCell`/`ReadWorksheet`/`ReadToc`/`ReadSection`/
+    `ReadVariables`/`WatchVariable`/`UnwatchVariable`), with zero networking
+    or JSON-RPC framing -- takes a `Worksheet*`/`Variablespane*`, returns
+    `nlohmann::json`, throws `McpToolError` for a bad tool name/arguments.
+    `McpServer` is purely the HTTP+JSON-RPC transport around it. This mirrors
+    why `TableOfContents`'s own structure-walking isn't reused as-is: its
+    public surface (`GetCell(displayedIndex)`, filtered by the TOC's own
+    search box/depth setting) is display-oriented, not a clean data API, so
+    `McpTools::ReadToc()` just re-walks `OnList(tree)` filtering
+    `GroupCell::IsHeading()` directly instead -- the same primitive
+    `TableOfContents::UpdateStruct()` itself uses.
+  - **Why "read-only except two things" and not stricter**: the maintainer
+    explicitly approved `watch_variable`/`unwatch_variable` (adding/removing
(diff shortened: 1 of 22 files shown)


gunterkoenigsmann:
"Compile using minGW" has failed on every push to this branch so far (`a08234c`, `457dfe7`, `1ef0e67`), but it's not this PR's doing: in every case the *only* failing test is `wxmaxima_version_string`, a pre-existing, intermittent failure on the minGW Windows runner that's been tracked separately since 2026-08-15 and reproduces the same way on unrelated pushes (see `AGENTS.md`'s "`wxmaxima_version_string` intermittent CI failure" entry for the ongoing investigation) -- every other job on this PR, including `compile_2204`, MSVC and the sanitized build, is green.

Re-ran the failed job once (run `34036922286`) since this test is known-flaky rather than deterministically broken; will keep an eye on it.

---

PR #2292, opened 2026-09-06
Give the AI worksheet context cursor/error awareness and output caps

gunterkoenigsmann:
## Summary

Three related gaps in the MCP server / AI chat sidebar's worksheet context (#2290, #2291), each raised directly during review of that work:

- **No notion of "the current cell" or "the cell above the cursor that errored."** An AI had no way to answer either question. Added `is_current`/`has_error` to `list_cells`/`read_cell`, and matching inline markers (`"(CURRENT CELL...)"`/`"(THIS CELL HAS AN ERROR)"`) in the plain-text worksheet dump the AI chat sidebar sends — it has no tool-calling yet, so JSON fields alone wouldn't reach it.
  - Found and fixed a related bug in the same pass: `AiChatSidebar::BuildContextSnapshot()`'s truncation for its 8000-character context limit truncated from the *start*, which would silently strip the new `(CURRENT CELL...)` marker out of any worksheet long enough to need truncating in the first place — exactly the worksheets where a user is most likely to ask about their current cell. Fixed by centering the kept window on the marker. Verified live with a real ~15KB synthetic worksheet: confirmed via temporary logging that the truncated ~8KB snapshot still contained both the marker and the last cell's distinctive content.
- **`read_variables` gave no way to tell "Maxima hasn't answered this query yet" apart from "this variable is undefined."** Both look identical (an empty value). Added a `maxima_busy` flag (derived from `Worksheet::GetWorkingGroup()`/`GetEvaluationQueue()`).
- **`read_cell` had no cap on a single cell's output at all**, and `read_worksheet`/`read_section` only capped the *aggregate* text, not each cell's own contribution — one huge cell's output (a large matrix, a long list, ...) could dominate or even crowd out other cells' content in the response (worse: since that aggregate cap also truncates from the start, a huge cell early in the document could crowd out the current-cell/error markers above too). Added `output_length`/`output_from_end` arguments to `read_cell` so an AI can ask for a preview or specifically the tail (e.g. "did this converge"), plus a smaller per-cell preview cap in the worksheet/section dumps with an inline pointer back to `read_cell` for the full text.

Full root-cause writeups for all three, including the live verification methodology, are in `AGENTS.md`.

## Test plan

- [x] `test_McpTools` — 68 assertions across 5 test cases (up from 53/4), covering `is_current`/`has_error` on `list_cells`/`read_cell`/`read_worksheet`, `maxima_busy`, and the `output_length`/`output_from_end`/`output_truncated` capping behavior
- [x] Full clean rebuild, zero new warnings
- [x] `check-pot-coverage` passes; the one changed translatable string (the AI chat sidebar's system prompt, now explaining the markers) is reflected in `wxMaxima.pot`
- [x] Live end-to-end verification in a real Xvfb session: a synthetic long worksheet, cursor moved to the last cell, confirmed the sent context snapshot correctly retained the current-cell marker despite truncation
- [ ] Full broad `ctest` regression suite was still running at push time; will report back if it turns up anything unrelated

---

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -703,6 +703,42 @@ a local TCP socket.
     (Gemini) API key..." Fixed by rewording to "Get an API key for %s...",
     which sidesteps the a/an agreement entirely rather than trying to track
     which of the four provider names needs which article.
+  - **Follow-up (2026-09-06): "do we need to hardcode the model names? they
+    tend to change over time"** -- a fair challenge, and the user's own
+    follow-up ("model auto detection feels like hunting a mechanism that
+    catches outdated strings with a mechanism that can get outdated") is
+    exactly why this wasn't turned into a live "fetch the model list from
+    each provider's API" feature: that mechanism would itself need
+    maintaining against four APIs just to answer a question a plain link
+    to each provider's own docs already answers, permanently, for free.
+    Two changes instead: (1) `AiProviderDefaultModel()`'s Anthropic entry
+    now uses `claude-3-5-sonnet-latest`, that provider's own "rolling"
+    alias, instead of the dated snapshot `claude-3-5-sonnet-20241022` it
+    used to be -- OpenAI/Google/Qwen's defaults (`gpt-4o-mini`/
+    `gemini-1.5-flash`/`qwen-plus`) were already un-dated in this same
+    sense, so only Anthropic's needed changing; this only pushes the
+    staleness problem up one level (from "this exact snapshot got retired"
+    to "this whole model line got superseded"), which a plain string
+    constant genuinely cannot solve by itself. (2) A new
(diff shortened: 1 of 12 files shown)


gunterkoenigsmann:
"Compile using minGW" failed on this PR's head (`f8a75cf`), but not because of anything in this PR: the only failing test is `wxmaxima_version_string`, a pre-existing, intermittent failure on the minGW Windows runner tracked separately (see `AGENTS.md`'s "`wxmaxima_version_string` intermittent CI failure" entry) — same known root cause already flagged on #2290, reproducing here identically. Every other check on this head, including `compile_2204`, MSVC, the sanitized build, and the rest of minGW's own 133 other tests, is green.

Re-ran the failed job once (run `34052154755`).

---

PR #2293, opened 2026-09-06
Add an MCP search_cells tool to find cells by text or regex

gunterkoenigsmann:
## Summary
- Adds `search_cells`, a new read-only MCP tool that finds every cell whose input and/or output contains a given plain-text substring (default, case-insensitive) or a POSIX extended regular expression (`"regex": true`), so an AI can jump straight to the relevant cell(s) of a large worksheet instead of reading everything via `list_cells`/`read_worksheet`.
- Reuses `wxRegEx`, the same regex engine `RegexSearch`/`FindReplacePane` already use elsewhere in this codebase, rather than adding a second dependency.
- Optional `case_sensitive` and `scope` (`"input"`/`"output"`/`"both"`) arguments narrow a search; results are capped to 50 matches (`truncated` flag) so a pathological pattern can't produce an unbounded reply.
- An invalid regex throws a proper `McpToolError` instead of matching nothing silently.

## Test plan
- [x] `test_McpTools` gained a new SCENARIO covering substring/regex/case-sensitivity/scope/invalid-pattern/the match cap (18 new assertions, 86 total in the file) — all passing.
- [x] Live-verified the transport end-to-end in Xvfb: started wxMaxima with the MCP server enabled, confirmed `search_cells` appears in `tools/list`, and exercised plain-substring, case-insensitive, and invalid-regex calls via `curl` against the real running server.
- [x] Full clean rebuild with zero new warnings; full `ctest` suite re-run with no regressions (one apparent failure, `testbench_simple.wxmx`, was confirmed to be an artifact of running that one test outside `xvfb-run` in isolation, not a real regression — it passes cleanly under `xvfb-run`).
- [x] Regenerated the POT and confirmed it is byte-identical in content (only the timestamp differs) — this tool's descriptions are AI-facing strings, not `_()`-wrapped user-facing UI text.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -739,6 +739,66 @@ a local TCP socket.
     having `ResetAllToDefaults()` call `AiProviderDefaultModel()` for all
     four, removing the second copy entirely; re-verified live in Xvfb
     that Options now genuinely shows `claude-3-5-sonnet-latest`.
+  - **Follow-up (2026-09-06): a new `search_cells` MCP tool, so an AI can
+    find a cell instead of reading the whole worksheet.** Raised by the
+    user directly ("would the AI profit from some regex search for input/
+    output?") -- a fair gap, since `list_cells`/`read_worksheet` were the
+    only way to find "which cell mentions X" and both mean reading every
+    cell yourself, exactly the friction the current-cell/error markers
+    above already exist to reduce. Added `McpTools::SearchCells()`: a
+    `pattern` argument, a plain case-insensitive substring by default, or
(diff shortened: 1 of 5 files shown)


gunterkoenigsmann:
"Compile using minGW" failed on this PR's head (`ef65b6e`), but it's the pre-existing, already-documented `wxmaxima_version_string` flake (see `AGENTS.md`'s long-running investigation into it) — 133/134 tests passed, only that one unrelated test failed with "Required regular expression not found." Nothing in this PR's diff (`search_cells`/`McpTools`) touches CI config, stdio redirection, or anything else in that failure's vicinity. Re-running the job once.

---

---

Issue #2274, opened 2026-08-15
Windows Dark Mode affects only the worksheet


gunterkoenigsmann:
### Discussed in https://github.com/wxMaxima-developers/wxmaxima/discussions/2273

<div type='discussions-op-text'>

<sup>Originally posted by **LM13700** August 14, 2026</sup>
Hello everyone.

According to the wxMaxima 26.07.1 release notes, it's possible to set the user interface to dark mode on Windows.
I tested that on both 26.07.01 and 26.08.0, but I can't get it to work.

What I did:
- Configure Maxima -> Style -> Appearance: change to `Dark`
- Only the worksheet is in dark mode, as described by the release notes
- Close wxMaxima and open it once again: still only the worksheet is in dark mode; the interface is `Light`

I tried running the program as an administrator, but no luck. Am I doing something wrong?

Best regards,
Mateusz</div>

gunterkoenigsmann:
Found a mechanistically precise root cause by reading wxWidgets 3.3's own MSW source — but I want to be upfront about a real limitation: this sandbox only has wxWidgets 3.2 installed, so the `wxCHECK_VERSION(3, 3, 0)` code path can't be compiled or run here at all. The fix below is well-evidenced from source reading, not verified on real Windows. It would help a lot if someone can confirm on Windows once this lands.

**The mechanism:** `main.cpp`'s `OnInit()` already had a comment explaining `ApplyAppearanceToApp()` (which calls `wxTheApp->SetAppearance()`) needs to run "before the first top-level window is created" for Windows to honor it. But the actual *first* thing `OnInit()` did, several hundred lines earlier, was construct the (normally hidden) log window: `m_logWindow = new wxLogWindow(...)`.

`wxLogWindow`'s constructor unconditionally does `m_pLogFrame = new wxLogFrame(...)` — a real `wxFrame` — regardless of the `show` argument; only the subsequent `Show()` call is conditional (confirmed by reading `src/generic/logg.cpp` directly). A `wxFrame` registers itself in wxWidgets' global `wxTopLevelWindows` list at *construction*, not at `Show()` time.

wx 3.3's MSW `wxApp::SetAppearance()` (`src/msw/darkmode.cpp`) opens with:
```cpp
if ( !wxTopLevelWindows.empty() || gs_appMode != AppMode_Default )
    return AppearanceResult::CannotChange;
```
[...]

PR #2294, opened 2026-09-06
Test wxmaxima_version_string's Windows CI flakiness under RUN_SERIAL

gunterkoenigsmann:
## Summary

This is a real experiment, not a fix — the underlying bug behind `wxmaxima_version_string`'s intermittent failure on the minGW Windows CI runner (documented at length in `AGENTS.md`) is still unsolved.

Prior sessions ran multiple Wine-based investigations against this test and disconfirmed several concrete mechanisms one at a time: a struct-copy-vs-`_dup2()` stdio-redirection theory, a shared-handle-after-close theory in `_dup2()` itself, and (closest to this PR) a synthetic two-process concurrent-execution repro that failed to reproduce the bug even under simulated contention. None of that Wine-based work ran under *real* Windows scheduling contention — the actual CI job runs this test as part of `ctest -LE "unittest|needs_posix" -E "multithreadtest" --repeat after-timeout:2 -j 2`, i.e. genuinely concurrently with another test process, on a shared GitHub Actions Windows runner (background load, real multi-core scheduler, antivirus real-time scanning, ...) that this sandbox's Wine environment cannot faithfully reproduce.

Building a full wxWidgets-for-MinGW cross-compile to get closer to that under Wine was judged too large an undertaking for a lead that a cheaper synthetic repro already failed to confirm. Instead, this sets `RUN_SERIAL TRUE` on just this one test, so CTest never schedules it concurrently with anything else regardless of `-j`. This tests the contention theory directly against the one environment that has ever actually reproduced the bug, instead of against another simulation of it:
- If the failure stops recurring across several real CI runs with this in place, that's genuine evidence contention is a necessary ingredient (worth then hunting for what shared state two concurrent `wxmaxima --version` processes could actually contend over).
- If it still fails, that rules out simple ctest-level contention cleanly, with no build effort spent either way.

`wxmaxima_version_returncode` (the sibling test that only checks the exit code, and has never been observed to fail) is deliberately left untouched — this experiment targets only the test that actually exhibits the bug.

**This should be reverted if a future session confirms it made no difference** — it's here to gather evidence, not presented as a fix.

## Test plan
- [x] Confirmed via `ctest --show-only=json-v1` that `RUN_SERIAL: true` is actually applied to `wxmaxima_version_string`.
- [x] Both `wxmaxima_version_string` and `wxmaxima_version_returncode` still pass locally under `xvfb-run` (this sandbox is Linux — the bug itself only manifests on the Windows CI runner, so this can't be verified to fix anything locally; only that the CMake change itself is inert here).
- [ ] Not verifiable here: whether this actually changes the failure rate on the real Windows CI runner — that needs several real CI runs to accumulate evidence either way.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1512,6 +1512,29 @@ a local TCP socket.
     **Revert this single `RUN_SERIAL TRUE` if a future session confirms it
     made no difference** -- it is an experiment to gather evidence, not a
     fix, and should not linger indefinitely presented as one.
+    **First two real data points (2026-09-06, PR #2294, commit `54598a7`):
+    both of the first two real CI runs with `RUN_SERIAL TRUE` in place
+    still failed with the exact same symptom** ("Required regular
+    expression not found", 133/134 tests passed, same single test). This
+    is meaningful, if not yet conclusive: simple ctest-level self-
+    concurrency (this test racing some *other* ctest job for CPU/scheduler
+    time within the same `-j 2` invocation) does not look sufficient on
+    its own to explain the failure, since removing exactly that kind of
+    contention for this one test didn't stop it from failing twice in a
+    row. Two important caveats before concluding contention is irrelevant
+    entirely: (1) `RUN_SERIAL` only keeps *this* test from running
+    concurrently with anything else -- it does nothing about contention
+    for the *machine's* resources in general (another GitHub Actions
+    Windows runner's own background load, antivirus scanning, etc. are
+    unaffected), so this doesn't rule out contention as a class, only
+    ctest's own internal `-j 2` scheduling specifically; (2) two data
+    points is still a small sample against a failure this reports as
+    "essentially every push" -- worth accumulating more real CI runs
+    before drawing a firm conclusion. Do not spend further Wine-repro
+    effort chasing plain ctest-level contention specifically based on this
+    -- that narrow mechanism now has two real, direct data points against
+    it, which outweighs the earlier from-first-principles Wine simulation
+    that failed to reproduce it either way.
 
 - **System tray icon (`src/TrayIcon.{h,cpp}`, GH #2286) -- mirrors the busy
   status, gated entirely by `wxUSE_TASKBARICON`.** The maintainer's own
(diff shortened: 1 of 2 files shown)


gunterkoenigsmann:
First real data point for this experiment: "Compile using minGW" failed on this PR's head (`54598a7`) with the usual `wxmaxima_version_string` symptom ("Required regular expression not found", 133/134 passed) — **despite `RUN_SERIAL TRUE` being in effect**. So a single concurrent `ctest -j 2` run isn't sufficient on its own to explain this failure; something else is still (at least also) required.

This is one data point, not a conclusion — the test's pre-existing failure rate is high enough ("essentially every push since 2026-08-15" per `AGENTS.md`) that a genuine contention effect could still exist but only be one contributing factor among several, or contention with the *other* concurrently-running test in the `-j 2` batch could still matter even with this one test serialized against everything else (RUN_SERIAL prevents this test from running concurrently with anything, but doesn't change what else was running around the same wall-clock window). Will keep watching for further runs before drawing a conclusion either way. Re-running the job once now.

---

---

gunterkoenigsmann:
Second data point: the re-run (same commit) failed with the identical `wxmaxima_version_string` symptom again. Two consecutive real CI runs with `RUN_SERIAL TRUE` in place, both still failing — meaningful evidence against plain `ctest -j 2` self-concurrency being the (sole) cause, though not yet conclusive (still a small sample against a failure reported as "essentially every push," and `RUN_SERIAL` only removes *this test's* internal ctest scheduling contention, not machine-level contention in general). Recorded in `AGENTS.md` (pushed as `45dc8ff`) so this doesn't need re-deriving. Not reverting the experiment yet — will keep watching for further runs.

---

---

gunterkoenigsmann:
Third data point, on a genuinely separate commit (`45dc8ff`, not a re-run): failed identically again. Three for three real CI runs with `RUN_SERIAL TRUE` in place, all failing the same way — closing out this specific lead as reasonably disconfirmed (pushed as `5530cad`). Leaving `RUN_SERIAL TRUE` in place since it's harmless and keeps the negative finding visible on the test itself, but no longer treating this as an open experiment. Future investigation on this bug should move to real Windows hardware with `rr`, or bisecting the GH #2274 startup-reordering commit directly — not another synthetic Wine repro of ctest-level contention, which this experiment has now ruled out as cheaply as it can be.

This PR is otherwise just the `RUN_SERIAL` line plus documentation — happy to leave it open for a maintainer to merge (it's a harmless, well-documented negative finding) or close, as preferred.

---

---

PR #2295, opened 2026-09-07
Add runtime-gated diagnostic tracing for the wxmaxima_version_string CI flake

gunterkoenigsmann:
## Summary

This is a diagnostic-only change, not a fix — `wxmaxima_version_string`'s intermittent failure on the minGW Windows CI runner (documented at length in `AGENTS.md`) is still unsolved after several Wine-based investigations that each disconfirmed one concrete mechanism (struct-copy vs. `_dup2()`, a shared-handle-after-close theory, ctest-level `-j 2` self-concurrency via `RUN_SERIAL`, and — checked directly this session — the theory that this only fails on PRs, which it doesn't).

Every prior Wine repro tested a bare `fprintf(stdout, ...)` write, never the *actual* code path `--version` uses: `wxMessageOutput::Get()->Printf(...)` through a `wxMessageOutputStderr` object, with a real `cmdLineParser.Parse()` call (`wxCmdLineParser`) running in between. Since GitHub Actions' `windows-latest` runner already is real Windows — the actual gap is its non-interactivity, not a lack of a genuine target — this has the binary record its own evidence instead of running yet another simulation.

**What's added**, all in `src/main.cpp` (Windows-only):
- `StdioDebugLog()`/`DescribeStdHandle()`/`DescribeStream()`: small helpers using raw Win32 I/O (`CreateFileW`/`WriteFile`), never CRT stdio, so tracing a stdio bug can't itself be perturbed by that same bug. Entirely inert unless `WXM_STDIO_DEBUG_LOG` names a log file — a single cached env-var check otherwise, in every normal build and every other CI job.
- Checkpoints bracketing the full chain: `BindStdStreamToParent()` (start/end, per stream), `RedirectStdioToParent()`, right after `SetHandleInformation(..., HANDLE_FLAG_INHERIT, 0)`, around `wxMessageOutput::Set(...)`, around `cmdLineParser.Parse()`, and entering the `-v` branch plus right after its `Printf()` call.

**CI wiring**, in `.github/workflows/compile_windows.yml`:
- `WXM_STDIO_DEBUG_LOG` is set only as an `env:` on the "Run integration tests" step (not job-wide), so the log only covers the one ctest invocation that reproduces this.
- A new "Upload stdio debug log" step follows it, uploading the resulting file as a build artifact (`if: always()`, `if-no-files-found: ignore` — a green run's log, showing nothing anomalous, is itself a useful data point).

No behavior change for macOS, Linux, or any Windows build/run that doesn't set `WXM_STDIO_DEBUG_LOG`.

## Test plan
- [x] Reviewed the full diff by hand for syntax/type correctness (no MinGW+wxWidgets cross-build available in this sandbox to fully compile-check) — in particular fixed a `wxString::Format` `%s`/`char*` width mismatch that could otherwise assert on a Unicode build, and removed an unconditional `fflush(stdout)` from an earlier draft since it would have been a real (non-diagnostic) behavior change riding along with what must stay strictly inert.
- [x] Confirmed the artifact-upload step's relative path resolves correctly against `${{ github.workspace }}`, matching where the traced process's `cd wxMaxima && cd build ... cd .. && cd ..` sequence leaves it.
- [ ] Not verifiable here: the actual captured trace content — that needs a real run on the Windows CI runner, which reproduces this failure on essentially every push.

See `AGENTS.md`'s `wxmaxima_version_string` entry for the full investigation history and the reasoning behind each design choice in this change.

---

### DIFF
--- a/.github/workflows/compile_windows.yml
+++ b/.github/workflows/compile_windows.yml
@@ -271,12 +271,33 @@ jobs:
         # unzip tests, which cannot run on Windows) are excluded; timeout-minutes
         # is a hard ceiling.
         timeout-minutes: 20
+        # WXM_STDIO_DEBUG_LOG: diagnostic tracing for the still-unexplained
+        # wxmaxima_version_string flake (see AGENTS.md's long investigation --
+        # this env var is the only thing that turns it on; main.cpp's tracing is
+        # a no-op without it, in every other build and every other CI job).
+        # Scoped to just this step, not job-wide, so the log stays small and
+        # only covers the one ctest invocation that actually reproduces the bug.
+        env:
+          WXM_STDIO_DEBUG_LOG: ${{ github.workspace }}\wxm_stdio_debug.log
         run: |
             cd wxMaxima
             cd build
             ctest -LE "unittest|needs_posix" -E "multithreadtest" --repeat after-timeout:2 -j 2 --output-on-failure
             cd ..
             cd ..
+      - name: Upload stdio debug log
+        # Only produced/useful when the step above actually failed -- see the
+        # WXM_STDIO_DEBUG_LOG env var on that step. if: always() (not
+        # if: failure()) because upload-artifact itself already no-ops cleanly
+        # when the file doesn't exist, and this keeps the step's own log
+        # readable ("no files were found" is more informative on a green run
+        # than never seeing this step at all).
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: wxm-stdio-debug-log
+          path: wxm_stdio_debug.log
+          if-no-files-found: ignore
       - name: Upload the unsigned installer for SignPath
         # SignPath does NOT accept an uploaded file directly: its GitHub
         # connector downloads the GitHub Actions artifact itself and checks the
--- a/.github/workflows/compile_windows.yml
+++ b/.github/workflows/compile_windows.yml
@@ -285,6 +285,22 @@ jobs:
             ctest -LE "unittest|needs_posix" -E "multithreadtest" --repeat after-timeout:2 -j 2 --output-on-failure
             cd ..
             cd ..
+      - name: Print stdio debug log
+        # Echoes the trace straight into this job's own console output, in
+        # addition to the artifact upload below -- a downloading agent (e.g.
+        # an AI coding assistant investigating this from a sandboxed session)
+        # may have no route to actions-results.blob.core.windows.net (the
+        # artifact's actual storage backend) even though it can read job logs
+        # via the GitHub API; this step makes the trace reachable either way.
+        if: always()
+        run: |
+            if (Test-Path wxm_stdio_debug.log) {
+              Write-Output "----- wxm_stdio_debug.log -----"
+              Get-Content wxm_stdio_debug.log
+              Write-Output "----- end wxm_stdio_debug.log -----"
+            } else {
+              Write-Output "wxm_stdio_debug.log was not produced."
+            }
       - name: Upload stdio debug log
         # Only produced/useful when the step above actually failed -- see the
         # WXM_STDIO_DEBUG_LOG env var on that step. if: always() (not
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1604,13 +1604,27 @@ a local TCP socket.
     harmless either way -- this test is fast and gains nothing from
     parallelism -- and the finding itself is worth keeping visible on the
     test), but stop describing it as an open experiment still gathering
-    data; it has its answer. The next real lead, if this is picked up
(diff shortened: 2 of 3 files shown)


gunterkoenigsmann:
`Compile using minGW` failed on the first run (commit `2aad5ab`) — but this is the **expected**, pre-existing `wxmaxima_version_string` flake this PR exists to diagnose, not a regression from this diagnostic-only change: `99% tests passed, 1 tests failed`, `wxmaxima_version_string (Failed)`, "Required regular expression not found" — the exact, long-documented signature (see `AGENTS.md`). The `wxm-stdio-debug-log` artifact uploaded successfully (5525 bytes), confirming the instrumentation itself works end-to-end and fired on the very run that reproduces the bug.

However, I couldn't actually read that artifact from this session: its storage backend (Azure Blob Storage, not `github.com` itself) is unreachable from this sandbox's network egress policy, even though the GitHub API call that hands back its (correctly-signed) download URL works fine. Pushed `0dfb7ba` to add a second, verified-reachable path to the same data: a step that echoes the log straight into the job's own console output (readable via the same GitHub API calls already used to pull this failure's test summary), in addition to the artifact upload.

This will need one more CI run on `0dfb7ba` to actually capture and read the trace — the real payoff of this PR. No action needed from a reviewer; documenting this here per the CI-failure protocol since it's a failure on a PR I opened.

---

gunterkoenigsmann:
`compile_latest_and_test` (Ubuntu) failed on `0dfb7ba` too, but on `lisp_mode` — unrelated to this PR, which only touches `main.cpp`'s Windows-only (`#ifdef __WXMSW__`) code and `.github/workflows/compile_windows.yml`. `lisp_mode.wxm` is a known intermittent Linux/Maxima-protocol-timing test (see the `m_configCommands` entry in `AGENTS.md`), nothing this diff can affect. A re-run attempt was rejected with "This workflow is already running" (the sibling `compile_and_test_sanitized` job in the same workflow run was still in progress) — will re-run the failed job once that finishes.

---

gunterkoenigsmann:
The instrumentation paid off — real trace captured and read (via the console-print fallback from `0dfb7ba`, since the artifact ZIP itself stays unreachable from this sandbox). Full write-up in `AGENTS.md`'s `wxmaxima_version_string` entry; summary here:

The trace for the actual failing process (the one that reached `--version`'s code path) is completely clean — the stdout fd/handle chain never changes across `_dup2()`, `SetHandleInformation`, `wxMessageOutput::Set()`, or `cmdLineParser.Parse()`, and the trace reaches "after Printf, before exit(0)" with no indication the write itself failed. This rules out every mechanism the instrumentation was built to catch (a silently-changing or dropped handle).

A genuinely new, unexpected fact surfaced as a side effect: `GetFileType()` on the initial `GetStdHandle(STD_OUTPUT_HANDLE)` reports `FILE_TYPE_CHAR`, not `FILE_TYPE_PIPE`, for *every* wxmaxima subprocess this ctest step spawns — not just the failing one. And `wxmaxima_version_string` turns out to be the *only* test in the whole suite whose pass/fail depends on content written through this specific `wxMessageOutput`/console-output code path (every other content-checked test's expected text comes from Maxima's own socket-based output, never touching this code at all) — so there's no comparable "same mechanism, but it passes" test to rule the FILE_TYPE_CHAR fact in or out.

This reframes the question from "something breaks wxMaxima's own stdio redirection" (it doesn't — the trace is rock-solid) to "does a write through a FILE_TYPE_CHAR-backed handle via this specific code path actually reach wherever ctest reads captured output from, for a WIN32-subsystem child process, on this toolchain." Concrete next steps (not yet attempted) are listed in `AGENTS.md`: instrumenting the write's own success/`errno`, a labeled `fflush()` experiment, and looking directly at how CMake/CTest's Windows child-process execution sets up stdio for a WIN32-subsystem executable specifically — that CTest mechanism, not wxMaxima's own code, now looks like the more likely place the real answer lives.

This PR's own goal (get a real trace, gated behind an env var, off a build that changes no behavior) is accomplished. Leaving it open/watched for continued CI health per the standing PR protocol, but the diagnostic payoff itself is captured.

---

gunterkoenigsmann:
`compile_windows` failed again on `100873b` — expected, since that commit only touches `AGENTS.md` (no code change) and the trace shows the same "entering -v branch" pattern as before, i.e. the same pre-existing `wxmaxima_version_string` flake this PR investigates, not a regression. `compile_ubuntu`'s earlier `lisp_mode` failure was a genuine flake too, confirmed by the re-run going green.

No further action needed here — this PR's own goal (ship inert, env-var-gated diagnostic tracing and capture a real trace) is complete, and the underlying flake itself remains a separate, larger investigation tracked in `AGENTS.md` with concrete next steps for whoever picks it up.

---

PR #2302, opened 2026-09-11
Add GitHub Models as a built-in AI Chat provider

gunterkoenigsmann:
## Summary

- Adds **GitHub Models** as a sixth built-in AI Chat provider (`AiProviderKind::GitHubModels`). It's GitHub's own official, OpenAI-compatible model-hosting API (`https://models.github.ai/inference/chat/completions`), authenticated with a plain GitHub personal access token via a standard `Authorization: Bearer` header — publicly documented at <https://docs.github.com/en/github-models>.
- Needed **zero new provider logic**: its request/response shape and auth header are byte-for-byte what `OpenAiCompatibleProvider` already implements for OpenAI/Qwen, so this is the same "OpenAI-compatible shape, different base URL/auth secret" pattern Qwen already established.
- Wired through every place a built-in kind needs: `Configuration` (new `AiApiKeyGitHubModels()`/`AiModelGitHubModels()` accessors, persistence, `ResetAllToDefaults()`), `ConfigDialogue.cpp`'s provider list and both of its per-kind switch statements (populate-on-open, apply-on-OK), and `AiChatSidebar::ReloadProviderFromConfig()`.

**Deliberately not implemented: real "GitHub Copilot Chat".** That assistant has no sanctioned third-party API at all — every community tool that reaches it does so by reusing an *approved editor's* OAuth client id (GitHub only allowlists Copilot's token-exchange endpoint for client ids it has specifically approved) and calling undocumented internal endpoints while impersonating that editor. That's a materially different (and riskier) thing than "no OAuth flow for a desktop app" — it risks the account being flagged/suspended under GitHub's Copilot terms — so it was intentionally left out. GitHub Models is the genuinely official, ToS-clean alternative that actually satisfies "talk to a GitHub-hosted model from wxMaxima."

## Test plan

- [x] Full clean rebuild (`ninja`), zero new warnings
- [x] `test_AiProvider` — new SCENARIO covering GitHub Models' request URL, Bearer auth header, and OpenAI-compatible request/reply shape (127 assertions / 9 test cases total, up from 116/8)
- [x] `test_ConfigRoundtrip` — unaffected, still passing (95 assertions)
- [x] Full non-batch/non-live-Maxima `ctest` suite under Xvfb — 145/145 passed, no regressions
- [ ] Not verified live against a real GitHub PAT (no network credentials in the build sandbox) — same caveat this codebase's other built-in AI providers (OpenAI/Google/Qwen) already carry

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1033,6 +1033,73 @@ a local TCP socket.
     (`wxUSE_SECRETSTORE` is off here, so the whole AI Chat tab stays hidden
     per its own gating) -- verified by code reading plus the unit test
     above, not a live screenshot.
+  - **Follow-up (2026-09-11): "can you implement a GitHub Copilot
+    connector?" -- researched directly with the user rather than just
+    coding it, since "GitHub Copilot" turned out to mean two genuinely
+    different things with very different risk profiles.** GitHub Copilot
+    Chat (the assistant bundled with a Copilot subscription) has **no
+    official third-party API at all**. The way every community tool that
+    reaches it (copilot.vim-style plugins, "copilot-api" proxies) actually
+    works: perform the GitHub OAuth device flow using *another product's*
+    client id (typically an approved editor's, e.g. VS Code's -- GitHub
+    only allowlists Copilot-token exchange for client ids it has
+    specifically approved, so a freshly-registered wxMaxima OAuth app would
+    authenticate fine via the device flow but be rejected at the next
+    step), then exchange the resulting token at an undocumented internal
+    endpoint (`api.github.com/copilot_internal/v2/token`) for a short-lived
+    Copilot token, then call another undocumented endpoint
+    (`api.githubcopilot.com/chat/completions`) while sending headers that
+    make the request look like it came from that editor. That is not "no
+    legitimate OAuth flow for a desktop app" (the situation already
+    documented for Anthropic/OpenAI/Google/Qwen/GitHub Models, all of
+    which still have a real pasted-API-key path) -- it is impersonating an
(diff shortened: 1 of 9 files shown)


gunterkoenigsmann:
Both current red checks are pre-existing and unrelated to this PR's diff — checked, not assumed:

- **`compile_latest_and_test`** fails on the POT-coverage check: `locales/wxMaxima/wxMaxima.pot` is missing several AI-Chat-related strings (`"(Custom)"`, `"API style:"`, `"Add custom provider..."`, `"Quick fill:"`, ...). This is already true on `main` — `git diff origin/main HEAD -- locales/wxMaxima/wxMaxima.pot` is empty, and this PR adds no new translatable strings at all (`AiProviderKindName`'s new `"GitHub Models"` case uses `wxS()`, not `_()`). Pre-existing drift from an earlier AI Chat sidebar change, not something this PR introduced or should fix inline (regenerating the POT here would also sweep in a large amount of unrelated translation drift — see `AGENTS.md`'s translations section on why that's a separate cleanup, not a bugfix commit).
- **`Compile using minGW`** fails with the long-standing `wxmaxima_version_string` flake ("Required regular expression not found. Regex=[wxMaxima .*", 99% tests passed / 1 failed out of 134) — this has been failing on essentially every push since 2026-08-15 regardless of the change under test (a Windows console/ConPTY output-draining race around a short-lived GUI-subsystem process's exit; see the extensive investigation already recorded in `AGENTS.md`). Not caused by this diff, which touches no Windows-specific or startup code.

No action taken on either from this PR.

---

PR #2307, opened 2026-09-14
Fix a cmd.exe quoting bug breaking wxmaxima_version_string on Windows CI

gunterkoenigsmann:
## Summary

`wxmaxima_version_string` has been failing on the minGW Windows CI job on essentially every push since 2026-08-15 (extensively investigated in `AGENTS.md`). A prior fix landed that routes the test's captured output through a real file instead of a live console handle, to sidestep a suspected Windows console/ConPTY output-draining race. That fix was directionally sound, but **it never actually got exercised** — its own `COMMAND` string had a self-inflicted `cmd.exe` quoting bug that made the test fail a different way, instantly, on every single run since.

- The `COMMAND` was `cmd /c "\"$<TARGET_FILE:wxmaxima>\" --debug ... && type ..."`. Once CMake unescapes the inner `\"..\"` and CTest's own Windows argument quoting wraps the whole (space-containing) string in an outer pair of quotes, `cmd.exe`'s `/c` argument starts with two adjacent quote characters — a well-known `cmd.exe` parsing trap.
- This matches the real CI failure signature exactly: `Required regular expression not found` followed by a literal `The syntax of the command is incorrect.`, at 0.01 sec, identically on every run — a rejected command line, not a race (a race wouldn't fail identically, instantly, every time).
- **Confirmed directly**, not just from reasoning about `cmd.exe`'s quoting rules: this sandbox has no real Windows, but does have Wine's `cmd.exe` reimplementation. Reproducing both the old and the fixed command string against `wine cmd /c "..."` pointed at a real `.exe` shows the old form failing to even recognize the target as a command, and the fixed form running it correctly.

## Fix

Drop the redundant inner quotes around `$<TARGET_FILE:wxmaxima>`. This branch only ever runs on this project's own CI, whose build path never contains a space, so the single outer pair of quotes CTest's own Windows quoting already adds is sufficient once nothing inside it starts with a stray quote character.

`AGENTS.md`'s long investigation entry is updated with the confirmed root cause and this fix, so a future session doesn't re-derive the (now superseded) ConPTY-race theory from scratch.

## Test plan

- [x] Reconfigured and rebuilt on Linux — the `if(WIN32)` branch is untouched there; confirmed the generated `else()` branch command is byte-for-byte identical to before
- [x] `ctest -R wxmaxima_version_string` passes locally under Xvfb
- [x] Reproduced the exact failure class (and the fix resolving it) against a real `cmd.exe` via Wine, pointed at an arbitrary real Windows executable
- [ ] Not yet independently re-confirmed against a real `compile_windows` CI run on this exact PR — awaiting that

---


gunterkoenigsmann:
`compile_latest_and_test`'s POT-coverage check failed here, but not because of this PR's own diff (it only touches `test/CMakeLists.txt` and `AGENTS.md`, no translatable strings). This branch's actual fork point predates two POT regenerations that have since landed on `main` (731eeab, 8b85d8e) — so its committed `wxMaxima.pot` was stale relative to its own tree. On top of that, `main`'s current HEAD had drifted stale *again* (the async-output/AI-connection-monitor work after 8b85d8e added more untranslated strings without another regen) — a separate, pre-existing issue on `main` itself, not something introduced by this PR.

Pushed a merge of `main` into this branch plus a fresh `wxMaxima.pot` regeneration (only the `.pot` committed, per this repo's convention — the resulting per-language `.po` drift was reverted). Verified locally against the exact `strip_pot`+`diff` comparison the CI step uses, before pushing.

---

---

Issue #2298, opened 2026-09-09
26.08.0 requires admin to install in Windows


jsto-dev:
This was not needed several releases ago and I see no technical reason why this should be required now.

daute:
The installer wasn't build as before (Crosscompiled using the https://github.com/wxMaxima-developers/Crosscompiled-Windows-installer repository), but now automatically (using Github actions. It might even be signed).

If necessary I can add a 'conventionally created' installer too. (My code-signing certificate is expired, it would be unsigned).

gunterkoenigsmann:
That's the classical security theatre thingy (https://xkcd.com/1200/): If no admin rights are needed any process can change the installed files whenever they want which clearly is a security issue. If you give the installer admin rights the installer can do everything. Microsoft says that is totally fine.

Issue #2196, opened 2026-08-04
A cell's first statement can be silently dropped and never sent to Maxima (found via tutorial_10Minutes flake)


gunterkoenigsmann:
## Summary

While root-causing an intermittent CI failure in `tutorial_10Minutes` (see PR #2195), I found something more serious than the test flake itself: **the first statement of a multi-statement input cell can be silently dropped by the client and never transmitted to Maxima at all** — no error, no warning, nothing sent over the wire.

## How this was found

`tutorial_10Minutes` runs `test/automatic_test_files/10MinuteTutorial.wxm`, which has this input cell:

```
assume(a > 0)$
integrate( 1 / (x^2 + a), x);
forget(a > 0)$
```

Intermittently (reproduced locally at roughly a 1-in-20 to 1-in-30 rate), the `integrate(...)` call asks the interactive "Is a positive or negative?" question — which it should never need to, since `assume(a > 0)` was sent first and Maxima answers that class of question from its assumption database without asking.

My first theory was that this was Maxima/GCL-side nondeterminism (e.g. hash-table iteration order affecting which internal branch `integrate()`'s algorithm explores). **That theory is wrong, and I want to flag that explicitly so nobody re-derives it and stops looking.** I disproved it directly with a packet capture of a live failing run:

```
tcpdump -i lo -w capture.pcap 'tcp portrange 49000-49999'
tshark -r capture.pcap -q -z follow,tcp,ascii,<stream>
```

The raw wxMaxima→Maxima wire transcript for the failing run shows:

```
integrate( f(var), var );
:lisp-quiet (setq $linel 64)
integrate( 1 / (x^2 + a), x);
```

`assume(a > 0)$` is **not in the transcript at all**. Not corrupted, not reordered — never sent. The client jumped straight from the previous cell's last command to this cell's *second* statement.

(For context: I went looking for transmission-layer explanations — corruption or packet reordering — because of a specific past incident recalled during this investigation, where a Lisp runtime's flush-without-waiting behavior hit an MTU-triggered code path that reordered packets while keeping their content correct. That turned out not to be the mechanism here either — the data is simply never written — but it's a good example of why checking the wire directly beats guessing.)

## What's confirmed vs. still open

- **Confirmed**: `assume(a > 0)$` never reaches `Maxima::Write()`.
[...]

Issue #2110, opened 2026-07-04
build tests populate  test/test_userdirs with bizarre folders that have weird names and weid permissions


jgmbenoit:
On an uptodate Debian/Sid environmen and with wxMaxima version 26.07.0, the test stage populates the `<BUILDDIR>/test/test_userdirs` with very bizarre folders that have weird names and weird permissions. This seems to affect the test itself as a closer look to  `<BUILDDIR>/Testing/Temporary/LastTest.log` shows.
For illustration, the command
```
ls -l <BUILDDIR>/test/test_userdirs/wxmaxima_performance3*
```
gives
```
<BUILDDIR>/test/test_userdirs/wxmaxima_performance3:
total 4.0K
drwxr-xr-t 3 jgmb jgmb 4.0K Jul  4 07:35 test-directory-cached
ls: cannot open directory '<BUILDDIR>/test/test_userdirs/wxmaxima_performance3'$'\n''       ': Permission denied
```
and the command
```
ls -ld <BUILDDIR>/test/test_userdirs/wxmaxima_performance3*
```
gives
```
drwxr-xr-x 4 jgmb jgmb 4.0K Jul  4 07:35  <BUILDDIR>/test/test_userdirs/wxmaxima_performance3
d--------x 2 jgmb jgmb 4.0K Jul  4 07:35 '<BUILDDIR>/test/test_userdirs/wxmaxima_performance3'$'\n''
```

gunterkoenigsmann:
Creating the folders is intentional: That solves the problem that ctest supports running tests in parallel but maxima chokes if one instance of maxima loads and compiles draw and saves the compilation results into the ${MAXIMA_USERDIR}/binary folder while another instance of maxima is instructed to load draw() and loads the half-saved compiled file. Assigning them strange permissions is - strange, though. My guess is that this is gcl-specific as I personally use sbcl and ctest seems to use the system's default permissions. Give me a week of time to investigate - but that seems to need investigation...

jgmbenoit:
I got the point. What is bizarre is the `d--------x` permissions and the `'$'\n'` in the names.

jgmbenoit:
What about launching a `maxima` job to load and compile the `maxima` libraries used during the tests  in one fixed `${MAXIMA_USERDIR}` place as a preamble to the tests ? This will allow to gain some compilation time and the test will be closer to a real situation (where multiple instances of `wxmaxima` are running).

[...]

Issue #72, opened 2011-04-20
Numpad dot chatter


andrejv:
Converted from [SourceForge issue 2783772](http://sourceforge.net/support/tracker.php?aid=2783772), submitted by m-h-s

Pressing the dot on the numeric keypad produces two dots inside a input cell.

If it's on a horizontal cursor, a double dagger (U+2021) and a dot are produced.

This is on WinXP,
wxMaxima 0.8.2 http://wxmaxima.sourceforge.net
Maxima 5.18.1 http://maxima.sourceforge.net

PR #2312, opened 2026-09-18
Add wxmaxima-cli.exe, a console-subsystem front end for Windows

gunterkoenigsmann:
## Summary

An executable's subsystem is a bit in its PE header, fixed at link time: a binary is GUI-subsystem or console-subsystem, never both. `wxmaxima.exe` has to be GUI-subsystem, or a console window appears beside every worksheet. That has a consequence no code inside `wxmaxima.exe` can fix — **cmd.exe does not wait for a GUI-subsystem process**, so it returns to the prompt before the child has printed anything, and `--version` / `--help` / `--batch` cannot behave the way a command-line tool is expected to. (This is the user-facing half of what `AGENTS.md`'s long `wxmaxima_version_string` entry chased from the test side.)

`wxmaxima-cli.exe` is a console-subsystem launcher: it starts `wxmaxima.exe` with its own standard handles, waits for it, and returns its exit code. Because *it* is console-subsystem, cmd.exe waits for it.

It deliberately links **none** of wxMaxima's own code and no wxWidgets — it needs neither, and staying dependency-free keeps it a few tens of kilobytes that link statically, which matters for a portable build. No changes to `main.cpp` were needed.

## Why a launcher rather than a shared core

A core DLL was considered first and rejected:

- Windows links wxWidgets **statically** (`-DwxWidgets_USE_STATIC=true`), so a DLL boundary would duplicate wx's global state — the `wxModule` registry, the `wxApp` instance, the art-provider table.
- It would need `__declspec(dllexport)` plumbing across ~200 wx-heavy classes plus templates like `CellPtr<>`, and add a second PE file to sign.
- It works directly against the portable/no-installer idea in #2298, whose point is fewer files and no dependencies.

Worth noting the subsystem bit is the *only* thing forcing a second binary — not code sharing. There is no separate diff binary to share a core with: `wxmxdiff` is a symlink to `wxmaxima` with `--diff` dispatch. Should this ever grow into a genuinely headless CLI that evaluates worksheets, the core should be shared through a CMake OBJECT library (the pattern `wxmTestApp`/`wxmFuzzApp` already use), not a DLL. That reasoning is recorded in `AGENTS.md`.

## Two details worth keeping

- **`STARTF_USESTDHANDLES` is load-bearing.** A GUI-subsystem child inherits no usable standard handles without it, even with `bInheritHandles=TRUE`. With it, `BindStdStreamToParent()` succeeds on its very first `GetStdHandle()` and never needs its `AttachConsole(ATTACH_PARENT_PROCESS)` fallback — and the same path works whether the launcher's own stdout is a console, a pipe or a file redirect.
- **The child's command line is the raw `GetCommandLineW()` tail, not a re-quoted `argv[]`.** Re-quoting is exactly the bug class that already cost this project a multi-session investigation, whose real root cause turned out to be one redundant pair of quotes.

## Test plan

- [x] Verified end-to-end **under Wine**, against a stand-in GUI-subsystem child reproducing `BindStdStreamToParent()` verbatim:
  - PE subsystem bits genuinely differ (2 = GUI vs 3 = CONSOLE, read out of the headers)
  - stdout reaches a pipe *and* a file redirect; stderr stays separate
  - `"C:\some path\my worksheet.wxmx"` survives as a single argument; embedded quotes survive
  - exit code propagates (42 in, 42 out)
  - the "installed as `wxmaxima.exe` myself" guard refuses rather than fork-bombing
- [x] Warning-clean with `-Wall -Wextra` on both `i686-` and `x86_64-w64-mingw32-g++`
- [x] Linux configure unaffected; the new target and test are correctly absent there
- [x] `check-pot-coverage` passes (the new file has no translatable strings by construction)
- [ ] **Not verified: against the real `wxmaxima.exe` on real Windows** — the new `wxmaxima_cli_version_string` ctest is what will say. Note it needs *none* of the route-through-a-file workaround its GUI-subsystem counterpart requires, which is precisely the property being added; if it ever does need that, the launcher has stopped doing its job.

## Not included

No SignPath or release-artifact changes are needed: binaries land in `build/src/` while CPack output goes to the build root, so the `wxMaxima/build/*.exe` globs are untouched. `wxmaxima-cli.exe` ships inside the NSIS installer, and signing the installer does not sign its contents — same as `wxmaxima.exe` itself today, so this introduces no new signing gap. The portable-ZIP idea for #2298 is independent and deliberately left out.

---


gunterkoenigsmann:
`compile_with_non_default_options` is red here, but **not because of this PR** — it is red on `main` itself, and has been for the last three pushes ([5016](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35271116956), [5015](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35270357134), [5014](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35269592704)). This PR only adds a `if(WIN32)` CMake target and a Windows-only source file; the failure is a Linux link error in code it does not touch:

```
/usr/bin/ld: undefined reference to `MyApp::OnFatalException()'
```

### Root cause

A guard mismatch between the declaration and the definition:

- `src/wxMaxima.h:818` declares it under `#if wxUSE_ON_FATAL_EXCEPTION && wxUSE_DEBUGREPORT` **only**
- `src/main.cpp:1198` defines it under `#ifdef USE_QA` **as well**

With `-DWXM_DISABLE_QA=YES` (which this job passes) `USE_QA` is undefined, so the method is declared `override` — and therefore emitted into the vtable — but never defined. Hence an undefined reference in `.data.rel.ro`, i.e. from the vtable rather than from any call site.

### Why the attempted fix in 8f129c3 didn't take

@daute — that commit was the right idea in the right place, but the macro name is off by a prefix. It used `#ifdef WXM_USE_QA`; the preprocessor symbol is **`USE_QA`** (`#cmakedefine USE_QA`, `src/BuildConfig.h.cin:19`). `WXM_USE_QA` is only the CMake option name and is never `#define`d, so the guard was false in *every* configuration — which deleted the declaration from the default build too, where `main.cpp` still defines the member, giving a different error. That's consistent with "That was not the problem..."; it *was* the problem, the guard just never evaluated true.

Worth noting the trap: `BuildConfig.h.cin` is inconsistent about this — `USE_WEBVIEW`, `USE_QA` and `USE_FRIBIDI` drop the `WXM_` prefix, but `WXM_USE_AI_TOOLS` keeps it. So the intuition "the macro is named like the option" is right exactly one time out of four.

The one-word fix (`WXM_USE_QA` → `USE_QA`, restoring 8f129c3 otherwise) is reproducing and verifying locally now; I'll put it up as its own PR against `main` and port it here so this PR can go green, rather than widening this one.

---

gunterkoenigsmann:
### `compile_latest_and_test` went red on `lisp_mode` — not this PR's, and demonstrably nondeterministic

Recording the evidence so it doesn't have to be re-derived.

**The failure** ([run 35311520621](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35311520621), commit `2da7c38`): `99% tests passed, 1 tests failed out of 233` — `15 - lisp_mode (Failed)`. Everything else, including `wxmaxima_version_string`, passed.

**Why it cannot be this PR's.** The whole diff is Windows-only plus documentation: `src/wxmaxima-cli.cpp` (new), a target and an install rule both inside `if(WIN32)` in `src/CMakeLists.txt`, a test inside `if(WIN32)` in `test/CMakeLists.txt`, and `NEWS.md`/`AGENTS.md`. On an Ubuntu runner that source file is never compiled and that test is never registered — there is no path by which it reaches a live-Maxima batch test. (The `src/wxMaxima.h` hunk now also present is the #2314 cherry-pick, a preprocessor guard inside `#ifdef USE_QA`, i.e. a no-op in the default configuration this job builds.)

**Why it's nondeterministic rather than simply pre-existing.** It is *not* red on the base — `compile_latest_and_test` passed on `main`'s head `201ff62` ([run 35271116956](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35271116956)). The stronger evidence is within the failing run itself: on that same commit, `compile_and_test_sanitized` ran its `run_tests` step to **success**. `lisp_mode` is not in that job's `-E` exclusion list (which covers only `tutorial_fittingEquations|multithreadtest|tutorial_toleranceCalculations|tutorial_displaying3DCurves|tutorial_solvingEquations|testbench_simple.wxmx`), so it really did execute there and pass. Same source, same test, two independent executions under `ctest -j 2`, opposite outcomes.

That matches `lisp_mode.wxm`'s documented history: `AGENTS.md`'s `m_configCommands` entry records an intermittent hang/failure in exactly this file, root-caused with `tcpdump` to the evaluation queue being advanced by one cell per stray main prompt. That specific cause was fixed — so this occurrence is either a second path to the same symptom or a relative of the still-open silent-statement-drop bug (#2196), which by its nature is invisible to almost every other test in the suite.

**Re-run:** not spent deliberately — pushing the #2314 cherry-pick started a full fresh run on `0bc3562e` on its own, which serves as it. I'll report back on that run rather than kicking another.

**Not touched:** I haven't skipped, excluded or retried `lisp_mode`, and won't. If it recurs it's worth its own investigation — a live-Maxima test that passes under ASan and fails without it is an unusual and probably informative shape.

The one thing on this PR still unproven anywhere is `wxmaxima_cli_version_string` against a real `wxmaxima.exe`; only the minGW job can say.

---

gunterkoenigsmann:
### minGW: two failures, one pre-existing and one genuinely mine (not yet fixed)

`Run integration tests` reported `99% tests passed, 2 tests failed out of 135`:

```
```

**#71 is not this PR's.** Its output is `The syntax of the command is incorrect.` — cmd.exe rejecting the command line before `wxmaxima.exe` ever runs. That is the failure `AGENTS.md`'s long entry describes, and the `cmd /c` de-quoting fix it documents as "pushed, awaiting the next run" is already on `main` (`test/CMakeLists.txt:716-720`) — **so that fix did not work.** This PR does not touch that test. Worth a separate look; the entry's own closing advice (read cmsys's actual Windows quoting in `Process_win32.c` rather than re-guessing from the symptom, or replace the nested `cmd /c` with a small checked-in `.bat` taking the exe path as an argument) is where I'd start.

**#72 is mine, and I have not fixed it — I could not root-cause it, so I have not guessed.** Here is where it stands.

It failed in **0.06 s with literally zero captured output**. `--output-on-failure` is set on that step, so the blank is real, not a reporting artifact.

What the log rules out, from evidence rather than reasoning:

- **The launcher ran and started the child.** None of its three error messages (`cannot determine my own location` / `refusing to launch a copy of myself` / `cannot start`) appears anywhere in the job log. So `ExeDirectory()` was non-empty, the self-spawn guard didn't fire, and `CreateProcessW()` **succeeded**.
- **Not a target-resolution or co-location problem.** No ctest `Unable to find executable`, and the build log shows `Linking CXX executable src\wxmaxima-cli.exe` — right beside `src\wxmaxima.exe`; neither target sets `RUNTIME_OUTPUT_DIRECTORY`, so the "look in my own directory" lookup is sound in the build tree.
- **Not the argument handling, as far as Wine can say.** I re-verified with a stand-in GUI child that echoes its own `GetCommandLineW()`, run from a foreign working directory to match ctest's `WORKING_DIRECTORY`: the child receives `"…\wxmaxima.exe" --debug --logtostderr --pipe --version`, complete and correctly quoted, through both a pipe and a file redirect.

The suggestive find, from `WXM_STDIO_DEBUG_LOG`: of the **88** traced wxmaxima processes in that run, **exactly one had `type=pipe` standard handles** — every other had `type=char`. That is consistent with it being this launcher's child, since nothing else in the suite forwards ctest's own handles via `STARTF_USESTDHANDLES`. Its trace:

```
BindStdStreamToParent(…285) start: handle=0x310 type=pipe → done: fd=1
BindStdStreamToParent(…284) start: handle=0x364 type=pipe → done: fd=2
BindStdStreamToParent(…286) start: handle=0x35c type=pipe → done: fd=0
after wxMessageOutput::Set: stdout fd=1 osHandle=0x230
after cmdLineParser.Parse() (result=0): stdout fd=1 osHandle=0x230
```

All three streams bound cleanly, `Parse()` succeeded — and it **never entered the `-v` branch**, i.e. it did not see `--version`. (Only two processes in the whole run reached `-v`; both had `char` handles.) So the handle forwarding — the part I'd have expected to be fragile — demonstrably worked, and the argument didn't arrive. But identifying that process by handle type alone is circumstantial, and that is exactly the gap worth closing before changing any behaviour.

**So I've pushed instrumentation, not a speculative fix** (`7681f63`). `CliDebugLog()`/`DescribeStdHandle()` mirror `main.cpp`'s `StdioDebugLog()` in every respect: same `WXM_STDIO_DEBUG_LOG` opt-in that exactly one CI step already sets, same raw `CreateFileW`/`WriteFile` with `FILE_APPEND_DATA` alone so concurrent processes append atomically, and never a byte to stdout, which belongs to the child. It records the raw command line, the std handle types, the extracted argument tail, the command line handed to `CreateProcessW()` and — the point — **the child's pid**, which will tie a specific traced wxmaxima process to this launcher instead of inferring it from handle types.

Verified both directions under Wine: variable unset → no log file at all and stdout byte-for-byte just the child's output; variable set → stdout unchanged, full trace in the file. Warning-clean under `-Wall -Wextra` on both `i686-` and `x86_64-w64-mingw32-g++`; `<cwchar>` included explicitly for `swprintf()` rather than relied on transitively, since the MSVC job builds this target too.

**This PR should not merge while #72 is red** — it's the one test that proves the feature does what it claims. I'll report what the next run's trace says.

(The earlier `lisp_mode` failure on the Ubuntu job is unrelated and covered in my previous comment.)

---

gunterkoenigsmann:
### Second Ubuntu run, a *different* test — and this one names its own cause

Following up rather than repeating myself: `compile_latest_and_test` went red again on `0bc3562e`, but not on `lisp_mode`. This time:

```
146 - printf_equations (SEGFAULT)
```

and unlike last time the log says why:

```
Error: Directory '.../build/test/test_envdirs/printf_equations/MAXIMA_USERDIR/.maxima'
       couldn't be created (error 2: No such file or directory)
Warning: Cannot create .../MAXIMA_USERDIR/.maxima, the directory maxima keeps
       configuration, user packages and caches
Error: Unable to initialize GTK+, is DISPLAY set properly?
```

That is the per-test environment directory failing to be created (ENOENT on a parent component) and the process then failing to reach the X display — under a single shared `xvfb-run -a` that the tests immediately before and after it used successfully. It's test-harness/runner environment, not worksheet logic; `compile_ubuntu.yml`'s own `ls -la test/test_envdirs  # check for issue #2110` line says this fragility is already known and filed.

So across two runs on this branch: two different live-Maxima GUI batch tests, each failing once, one of them with an explicit environment error. Against that: `compile_latest_and_test` is green on `main`'s head, and on the first of these two runs the ASan job ran the same suite on the same commit to success.

**I did check one hypothesis that pointed at my own diff before writing this off**, because a SEGFAULT deserves it: the cherry-picked `#ifdef USE_QA` guard wraps the declaration of `MyApp::OnFatalException()` — wxWidgets' *fatal-exception handler*. If that were compiled out here, a crash that used to be caught and turned into a debug report could surface as a raw SEGFAULT instead, which would fit the symptom suspiciously well. It isn't: this job configures with `cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DWXM_UNIT_TESTS=YES ..` and no `-DWXM_DISABLE_QA`, so `USE_QA` is defined, the guard is true, and the declaration is present exactly as before. The cherry-pick is a genuine no-op in this configuration.

No further action from me on the Ubuntu side, and I have still not spent a deliberate re-run — pushing `7681f63` started a third full run on its own, which is now in flight. That run's minGW job is the one I actually care about: it carries the first `wxmaxima-cli` trace, including the child's pid, which should finally settle `wxmaxima_cli_version_string`. I'll report what it says.

---

gunterkoenigsmann:
### The trace came back, and it settles it: the launcher is doing its job perfectly. The output is lost anyway.

**First, a correction to what I wrote above.** My previous comment's suggestive reading — that the pipe-handled child "never entered the `-v` branch, i.e. it did not see `--version`" — was **wrong**. It rested on identifying the child by handle type, and the pid now proves otherwise. Here is the real trace from `7681f63`:

```
[wxmaxima-cli pid 6256] start: raw GetCommandLineW()=[D:/a/.../wxmaxima-cli.exe --debug --logtostderr --pipe --version]
[wxmaxima-cli pid 6256] start: stdin handle=…358 type=pipe, stdout handle=…3a0 type=pipe, stderr handle=…314 type=pipe
[wxmaxima-cli pid 6256] argument tail=[--debug --logtostderr --pipe --version]
[wxmaxima-cli pid 6256] CreateProcessW: exe=[D:\…\wxmaxima.exe] cmd=["D:\…\wxmaxima.exe" --debug --logtostderr --pipe --version]
[wxmaxima-cli pid 6256] CreateProcessW ok, child pid 9168
[wxmaxima-cli pid 6256] child exited with code 0
```

and child `9168`'s own trace ends:

```
[pid 9168] entering -v branch: stdout fd=1 osHandle=…244, wxMessageOutput::Get()=…
[pid 9168] after Printf, before exit(0)
```

Every link holds. The argument tail is extracted correctly, the child's command line is correct and correctly quoted, the child parses it, reaches the version branch, completes `Printf()` and exits 0 — **and ctest captured zero bytes.** `ArgumentTail()` is exonerated by evidence now, not just by a Wine repro that couldn't reproduce the failure either way.

### This hands the `wxmaxima_version_string` investigation a real fact

`AGENTS.md`'s long entry on that bug has one central unexplained observation: every wxmaxima child sees `FILE_TYPE_CHAR` where a pipe was expected, and it speculated this was something about how CTest sets up stdio for a WIN32-subsystem child. This run settles part of that.

The launcher is an ordinary **console**-subsystem process. It sees its own three std handles as **`type=pipe`** — so CTest really does hand its direct children pipes, exactly as the textbook assumption said — and passes those same handles on via `STARTF_USESTDHANDLES`. The child's reported handle *values* are identical (`0x3a0`/`0x314`/`0x358`), so inheritance demonstrably worked. The GUI-subsystem child nonetheless reports those very same handles as **`type=char`**.

So the pipe→char transformation is tied to **the child being GUI-subsystem**, not to anything CTest does. It isn't stable either: in the previous run exactly one child reported `type=pipe`; in this run that same child reports `type=char`.

### What this means for this PR

I conflated two problems, and should say so directly. A console-subsystem launcher fixes the "**cmd.exe does not wait for a GUI-subsystem process**" half — that is now *proven*, not argued: it waits, and it propagates the exit code (0, from a child that really did run). It does **not** make `--version` output reach a *capturing* parent on this runner. `wxmaxima_cli_version_string` fails for the same unresolved reason `wxmaxima_version_string` does, and both fail in the same run:

```
```

**I have not "fixed" my test by routing it through a file** the way its sibling does, and I'd argue against doing so: that hides the one signal that will show when the underlying capture bug is actually solved. I've also not disabled or excluded it.

Pushed `6c38a32`, which corrects the `AGENTS.md` entry I wrote earlier today — it recorded the wrong guess, and leaving that standing would cost the next person real time.

**So this PR should not merge as-is**, and the call is yours:

1. **Hold** until the output-capture bug is understood — it now has a much sharper lead than it did this morning.
2. **Merge the launcher, mark the test `WILL_FAIL`** with a comment pointing at that bug, so the feature lands and the test flips to a failure the day it's fixed.
3. **Merge the launcher, drop the test**, tracking it on the capture issue instead.

I'd take (2): it keeps the coverage honest and self-announcing. But dropping or inverting a test is your decision, not mine to make unilaterally.

Worth noting the user-visible feature is likely unaffected by any of this — an interactive `cmd.exe` writes to a real console, not a captured pipe, and the waiting behaviour that users actually hit is the part that's proven. I left `NEWS.md` as written for that reason rather than loading release notes with an investigation caveat; say the word if you'd rather it were hedged.

---

PR #2313, opened 2026-09-18
Publish the portable Windows .zip to the release as well (GH #2298)

gunterkoenigsmann:
## Summary

Closes the practical half of #2298. The installer needs administrator rights for a legitimate reason — it writes to Program Files and registers file associations for every user — and this PR does **not** change that. It gives the users who cannot or will not elevate something else to download, which turned out to already exist and simply never be attached to the release.

`CPACK_GENERATOR` is `"ZIP;NSIS"`, so every CI run already builds a portable `.zip` of exactly the same tree. It is already verified, too: the existing *"Verify the package bundles the MinGW runtime DLLs"* step asserts the MinGW runtime is inside **that very archive**, and fails the job outright if no ZIP was produced. So this is a one-glob change to a package that was already being built and checked on every run.

## Why the unpacked tree actually works

Not assumed — checked in `src/Dirstructure.cpp`:

- `ResourcesDir()` starts from `wxStandardPaths::Get().GetExecutablePath()`, walks up out of `bin/`, and appends `share/` if present. No registry lookup, no hardcoded Program Files path.
- The package layout is exactly that shape: `<root>/bin/…` + `<root>/share/…`.
- The archive carries the whole `share/` tree, **including all 50 locale `.mo` files** (verified by unpacking a locally generated CPack ZIP) — so a portable copy is not a translation-less one, which is what I'd initially assumed a "portable build" would cost.

## Scope

- The installer is untouched and stays the recommended option.
- What the `.zip` does *not* do: register file associations or create Start Menu entries. That's inherent to not installing, and it's in the NEWS entry.
- Neither Windows package bundles Maxima itself — the *combined* installer is built in the separate `Crosscompiled-Windows-installer` repo (per `src/CMakeLists.txt`'s own comment, and matching @daute's comment on the issue). The ZIP is therefore no worse off than the installer in that respect, but it's worth being aware of when answering the issue.
- No SignPath change. A `.zip` is not a PE file; the executables *inside* it are unsigned exactly as they are inside the NSIS installer today, so this introduces no new signing gap. Whether to sign the inner executables is a pre-existing, separate question.

## Test plan

- [x] `.github/workflows/compile_windows.yml` parses as YAML, and the `files:` value comes out as the two newline-separated globs `softprops/action-gh-release` expects
- [x] Confirmed CPack writes finished packages to the **build root** (ran `cpack -G ZIP` locally: `build/wxMaxima-26.08.0-Linux.zip`, with no stray archives in `_CPack_Packages/`) — which is why the glob is deliberately non-recursive
- [x] Confirmed the package layout is `bin/` + `share/`, matching what `Dirstructure::ResourcesDir()` walks, and that all 50 locale `.mo` files are present
- [x] A missing ZIP cannot slip through silently: the existing verify step already hard-fails the job before `Publish release` is reached
- [ ] Not verifiable here: the actual attach on a real `Version*` tag, which only happens on a tagged run

---


gunterkoenigsmann:
`compile_with_non_default_options` is red here, but **not because of this PR** — this PR changes one workflow file, `NEWS.md` and a docs file, and contains no C++ at all. The failure is a Linux link error (`undefined reference to MyApp::OnFatalException()`) that is red on `main` itself, and has been for the last three pushes ([5016](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35271116956), [5015](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35270357134), [5014](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35269592704)).

Full diagnosis in #2312: https://github.com/wxMaxima-developers/wxmaxima/pull/2312#issuecomment-5725794397 — briefly, `MyApp::OnFatalException()` is declared in `src/wxMaxima.h` without the `USE_QA` guard that its definition in `src/main.cpp` has, so with `-DWXM_DISABLE_QA=YES` it lands in the vtable undefined.

A fix is being verified locally now and will go up as its own PR against `main`; I'll port it here so this PR can go green rather than waiting on the merge.

---

gunterkoenigsmann:
### `Compile using minGW` is red on `wxmaxima_version_string` — pre-existing, not this PR's

```
99% tests passed, 1 tests failed out of 134
```
with `The syntax of the command is incorrect.` — cmd.exe rejecting the command line before `wxmaxima.exe` ever runs.

**Why it isn't this PR's:** the whole diff is the `Publish release` step's `files:` glob, plus `NEWS.md`, a skill doc, and the #2314 cherry-pick in `src/wxMaxima.h`. The `Publish release` step is `skipped` on a non-release run, so this PR changes nothing that executes here. It doesn't touch `test/CMakeLists.txt` or anything `wxmaxima_version_string` runs.

**Positive confirmation rather than just an argument from the diff:** the same job on #2312 fails the *same* test with the *same* message, and #2312 shares none of this PR's changes. The two branches differ in every file except the shared cherry-pick, so the common factor is `main`.

**Worth flagging for a separate issue:** `AGENTS.md`'s long `wxmaxima_version_string` entry documents a root cause (a redundant inner pair of quotes in the `cmd /c` command line) and a fix as "pushed, awaiting the next `compile_windows` run". That fix **is** on `main` (`test/CMakeLists.txt:716-720`) and the test **still** fails with the identical error — so it did not work, and the entry's confident close should be read with that in mind.

I looked at whether I could fix it here and deliberately stopped: I can't validate a fix for it from this environment. Wine's `cmd` is a reimplementation that strips the outer quotes and runs the current form happily, so it cannot adjudicate real `cmd.exe`; and by the documented `cmd /c` quote rules (`cmd /?`, rule 2) the current form *should* already work, so the obvious `cmd /s /c` tweak is not clearly right either. Guessing from outside is exactly what that entry's history shows going wrong repeatedly, so I'd rather not add a fifth disproven theory. Not widening this PR with a speculative change to an unrelated test.

One thing that may make it moot: `wxmaxima-cli.exe` (#2312) exists precisely so a command prompt gets wxMaxima's output normally. Once that works, `wxmaxima_version_string` can invoke it directly and the `cmd /c … > file && type file` workaround — and this entire class of nested-quoting fragility — can go away. That's the more promising route than another round of quoting archaeology.

**Re-run:** not spent deliberately — the #2314 cherry-pick push started a fresh full run on `30d42ae` on its own. Keeping this PR watched until it's green and mergeable.

---

PR #2314, opened 2026-09-18
Fix the link error in builds with -DWXM_DISABLE_QA=YES

gunterkoenigsmann:
## Summary

`compile_with_non_default_options` has been red on `main` since 2026-09-17 ([5014](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35269592704), [5015](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35270357134), [5016](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35271116956)) with:

```
/usr/bin/ld: undefined reference to `MyApp::OnFatalException()'
```

`MyApp::OnFatalException()` is **declared** in `src/wxMaxima.h` under `#if wxUSE_ON_FATAL_EXCEPTION && wxUSE_DEBUGREPORT`, but **defined** in `src/main.cpp` under `#ifdef USE_QA` *as well*. With `-DWXM_DISABLE_QA=YES` it is therefore declared and never defined — and because it is an `override`, it does not simply go unused: it is emitted into `MyApp`'s vtable. That's why the linker points at `_ZTV5MyApp` rather than at any call site.

The fix adds the missing `#ifdef USE_QA` around the declaration — **the exact nesting `GenerateDebugReport()` already carries a few lines further down in the same class** (`src/wxMaxima.h:885`), which is the QA-only method `OnFatalException()` itself calls.

## On 8f129c3 / 201ff62

@daute — 8f129c3 was the right diagnosis in the right place; only the macro *name* was off. It guarded on `WXM_USE_QA`, which is the **CMake option** name and never becomes a preprocessor macro: `src/BuildConfig.h.cin` turns it into `#cmakedefine USE_QA`. So the guard was false in *every* configuration, which also removed the declaration from the default build — where `main.cpp` still defines the member, giving a different error. That's consistent with concluding "That was not the problem..."; it was the problem, the guard just never evaluated true.

The trap is real and worth knowing: `BuildConfig.h.cin` is inconsistent about the prefix — `USE_WEBVIEW`, `USE_QA` and `USE_FRIBIDI` drop `WXM_`, while `WXM_USE_AI_TOOLS` keeps it. So "the macro is named like the option" holds exactly one time in four.

## Test plan

Verified locally in **both** directions, since the previous attempt was broken precisely by only holding in one:

- [x] `cmake -DWXM_DISABLE_QA=YES …` — **reproduced the exact failure first** on unfixed code:
      `undefined reference to MyApp::OnFatalException()` at
      `.data.rel.ro._ZTV5MyApp[_ZTV5MyApp]+0x108` (without LTO the symbol is explicit, confirming it comes from the vtable), then **links cleanly** with the fix applied.
- [x] Default configuration (`USE_QA` defined) — still builds and links, i.e. the declaration is still present for `main.cpp`'s definition to match. This is the direction 8f129c3 broke.
- [x] Confirmed `USE_QA` is genuinely visible where the guard sits: `src/wxMaxima.h` includes `BuildConfig.h` directly (line 34), so this is not a repeat of the "guard on an undefined macro" failure.

Once this lands I'll merge it into #2312 and #2313, whose `compile_with_non_default_options` runs are red for this same pre-existing reason.

---


gunterkoenigsmann:
### The fix is confirmed on real CI; the one remaining red check is pre-existing

**`compile_ubuntu` is green on `6b3aff9`** ([run 35312886246](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35312886246)) — including `compile_with_non_default_options`, the job that has been red on `main` since 2026-09-17 and that this PR exists to fix. That's the real confirmation; my local reproduction predicted it, and CI agrees. The default-configuration jobs in the same run stayed green too, so the change is a no-op where `USE_QA` is defined — the direction 8f129c3 broke.

**`Compile using minGW` is red, and it is not this PR's:**

```
99% tests passed, 1 tests failed out of 134
```

with `The syntax of the command is incorrect.` — cmd.exe rejecting the command line before `wxmaxima.exe` runs. This PR's entire diff is a 15-line preprocessor guard in `src/wxMaxima.h`; it touches nothing that test executes.

Confirmed rather than argued: the same single failure, same message, same 134-test count appears on #2313 (whose diff is a workflow glob plus docs) and on `main` itself. Three branches with disjoint diffs, one shared failure — the common factor is the base.

Worth flagging while here: `AGENTS.md`'s long `wxmaxima_version_string` entry records a root cause (a redundant inner pair of quotes in the `cmd /c` command line) and a fix as "pushed, awaiting the next `compile_windows` run". That fix **is** on `main` (`test/CMakeLists.txt:716-720`) and the test still fails identically — so it did not work, and that entry's confident close should be read with that in mind. I deliberately did not attempt a follow-up fix from here: Wine's `cmd` is a reimplementation that runs the current form happily, so it can't adjudicate real `cmd.exe`, and by the documented `cmd /c` quote-stripping rules (`cmd /?`, rule 2) the current form *should* already work — meaning the obvious `cmd /s /c` tweak isn't clearly right either. That entry's history is a series of plausible mechanisms that didn't survive contact with the platform, and I'd rather not add a fifth. It deserves its own issue and someone who can iterate against real Windows.

**Re-run:** not spent — no push of mine has been needed here, and the failure is reproducible on three branches, so a re-run would only re-confirm it.

This PR is green on everything it can affect. Ready for review as far as I can tell.

---

PR #2315, opened 2026-09-18
Stop recopying all 50 translations on every build

gunterkoenigsmann:
## Summary

`ninja` could never report "no work to do", even on a completely unchanged tree. This is why.

`copy_mo_file_${LANG}_for_wxmaxima_local` copies each `.gmo` into `${CMAKE_BINARY_DIR}/share/locale/<lang>/LC_MESSAGES/wxMaxima.mo`, so `./wxmaxima-local` finds its translations without being installed. It was written as:

```cmake
add_custom_target(copy_mo_file_${LANG}_for_wxmaxima_local ALL DEPENDS <gmo>
          COMMAND ${CMAKE_COMMAND} -E copy <gmo> <dest>.mo ...)
```

A custom **target** is by definition always out of date: its `COMMAND` runs on every build, and its `DEPENDS` only *orders* it after the `.gmo` — it never makes the copy conditional. Only `add_custom_command(OUTPUT ...)` gets a real up-to-date check.

With one of these instantiated per language inside `foreach(POFILE ${TRANSLATION_FILES})`, **every** build spawned 50 redundant `cmake -E copy` processes, including a no-op one.

The fix names the destination `.mo` as a real `OUTPUT` and leaves the `ALL` target as a thin `DEPENDS` on it. The target keeps its name and its `ALL` tag.

## Measured, not inferred

Three consecutive passes over the 50 real targets, before and after:

| | pass 1 | pass 2 | pass 3 |
|---|---|---|---|
| before | 50 copies | 50 copies | 50 copies |
| after | 50 copies | **0** | **0** |

After the change, pass 2 ends in `ninja: no work to do.`

I also isolated the mechanism in a minimal CMake project containing both shapes side by side, to be sure the difference was the pattern itself and not something about this tree: the `add_custom_target`-with-COMMAND form re-ran on every no-op build, the `add_custom_command(OUTPUT ...)` form went quiet after the first.

## Verified behaviour-preserving

Both directions that matter:

- [x] `diff -r` of the pre- and post-change build trees' `share/locale` is **identical** — the same 50 files, byte for byte, in the same places.
- [x] `touch`ing a `.gmo` still re-runs *that* language's copy, and the build then settles again. The dependency tracking genuinely works rather than merely going quiet.
- [x] The target keeps its name and `ALL` tag, so `src/CMakeLists.txt`'s macOS bundling step — which copies this same directory into the `.app` and names these targets in its comment (GH #1711) — is unaffected, as is the `install(FILES ...)` rule, which uses the `.gmo` directly and was not touched.

## Not an instance of the same bug

`info/CMakeLists.txt`'s `add_custom_target(html ALL)` looks similar but is fine: it carries no `COMMAND` at all and is a pure aggregator over `build_*.html` targets, whose own `add_custom_command(OUTPUT ...)` rules are already output-driven. Pandoc does not re-run when the HTML is current. I checked this rather than assuming it, and it needed no change.

## Also here

A short `NEWS.md` entry (the file does carry build-side entries) and an `AGENTS.md` note, since "a COMMAND on an `add_custom_target` always re-runs, and `DEPENDS` does not change that" is exactly the kind of thing that gets reintroduced by someone reading the old shape as a working example.

---


gunterkoenigsmann:
`Compile using minGW` is red here, on the two known Windows stdio-capture tests — not on anything this PR changes:

```
99% tests passed, 2 tests failed out of 135
The following tests FAILED:
```

This PR touches `locales/wxMaxima/CMakeLists.txt`, `NEWS.md` and `AGENTS.md` — nothing in the path that captures a Windows process's stdout.

Why it isn't this PR's, from the base branch rather than from reasoning:

* `wxmaxima_version_string` is red on `main` right now. The most recent completed `compile_windows` run on `main` ([6ecc9c8](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35336663335), the #2313 merge) reports `1 tests failed out of 134` — `71 - wxmaxima_version_string (Failed)`. It has failed on essentially every push since 2026-08-15; that investigation is written up at length in `AGENTS.md`.
* `wxmaxima_cli_version_string` is new: it arrived with #2312, which is why `main`'s run above counts 134 tests and this one counts 135. `main`'s first run carrying it ([fecf7f1](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35336734143)) is still compiling, so I can't yet cite it as red there — but its cause is already established. The instrumentation added in #2312 shows the launcher doing its job perfectly: correct argument tail, correct quoting, `CreateProcessW ok`, the child reaches the `-v` branch and completes its `Printf()` and exits 0 — and ctest still captures zero bytes. It fails for the same unresolved reason as `wxmaxima_version_string`.

No fix exists to port. The underlying bug is open and unsolved: a GUI-subsystem child sees the inherited pipe handles as `FILE_TYPE_CHAR` while the console-subsystem launcher that passed it those same handle values sees them as `FILE_TYPE_PIPE`, so the pipe→char transformation is tied to the child's subsystem rather than to anything CTest does. That's the current lead, recorded in `AGENTS.md`.

I have not re-run the job: it would not be diagnostic, since the failure is present on the base branch rather than intermittent. And I have deliberately not routed either test through a file or disabled it — that would hide the one signal that will show when the capture bug is actually fixed.

---

PR #2316, opened 2026-09-18
Document the one-branch-per-change workflow

gunterkoenigsmann:
## Summary

Writes down the workflow you want, which wasn't recorded anywhere: **one branch per feature or bugfix, cut fresh from `main`, and a pull request once that feature is finished.**

It goes in `AGENTS.md` — the file every agent reads, and which `CLAUDE.md` imports — as a short section immediately after the intro, before `## Build System`, since it's process guidance that applies before any code is touched. There's also a one-line cross-reference from `Conventions & Standards`, next to the existing git bullet, for anyone who looks there first.

## Why it's stated with a reason rather than as a bare rule

Because the alternative has already cost time here, and a rule with its reason attached survives better than one without. `claude/wxmaxima-cell-footprint-km7ptf` was reused for ~21 successive PRs. Each PR's content reached `main` under different commit ids, so the branch itself was never fast-forwarded, and it drifted ~60 commits behind. The four commits it still carried *looked* like unmerged work; establishing that they were in fact already duplicated in `main` (`EvaluationStatus`, `maxima_connected`, `SetConnectionCheck`, `MarkCommandSent`, and the fixed `wxFlexGridSizer` row counts) had to be done symbol by symbol before anything could safely be pushed to it. A branch cut fresh from `main` per change cannot reach that state.

## Two consequences the section calls out

- **An agent's own session configuration may name a fixed development branch** and instruct it never to push elsewhere. That is in direct conflict with this workflow — and since a fresh session starts from that template again, the template is what needs changing, not just the one session. Worth writing down so the next agent recognises the conflict instead of silently reusing the fixed branch.
- **An unrelated fix found while a PR is open belongs on its own branch**, even when it is one line.

## Not included

No `NEWS.md` entry. `AGENTS.md`'s mandate asks for one for every non-trivial change, but this changes how agents work on the repository rather than anything a user of wxMaxima can observe, and `NEWS.md` is user-facing release notes. Happy to add one if you'd rather it were recorded there too.

Docs only — no code, no build files, no tests touched.

---


gunterkoenigsmann:
`Compile using minGW` is red here. Every other check on this PR is green.

```
99% tests passed, 2 tests failed out of 135
The following tests FAILED:
```

This is not this PR's failure, and it can't be: the only file this PR changes is `AGENTS.md`. Nothing here is compiled, linked or executed.

`wxmaxima_version_string` is red on `main` right now — the most recent completed `compile_windows` run there ([6ecc9c8](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35336663335)) reports that same test as its one failure. `wxmaxima_cli_version_string` arrived with #2312 and fails for the same unresolved reason. The details, and why no fix exists to port yet, are in [my comment on #2315](https://github.com/wxMaxima-developers/wxmaxima/pull/2315#issuecomment-5730057600) rather than repeated here.

No re-run: the failure is present on the base branch, so re-running would not be diagnostic.

---

Issue #2, opened 2011-04-20
font and/or locale problem with the debian package


andrejv:
Converted from [SourceForge issue 1151606](http://sourceforge.net/support/tracker.php?aid=1151606), submitted by lindi2

Hi,

I recently installed

http://prdownloads.sourceforge.net/wxmaxima/wxmaxima_0.5.4-1_i386.deb?download

and tried to run it in my debian sid system. A popup
with "Cannot set locale to en_GB is shown" and text is
unreadable. A screenshot is probably in order here:

http://iki.fi/lindi/wxmaxima-cannot-set-locale-to-en-GB.png

Locale information:
LC_CTYPE=fi_FI
LANG=C
LESSCHARSET=latin1

After generating the en_GB locales with
dpkg-reconfigure locales I no longer see the warning
but the text is still unreadable. Maybe I'm missing
some font? If yes, this should be set as a dependency.
Also, I'm running wxmaxima in a VNC server running on
my debian woody system.

Issue #1, opened 2011-04-20
Can't see any entries  in main window - [wxGTK2-2.4.2]


andrejv:
Converted from [SourceForge issue 1112642](http://sourceforge.net/support/tracker.php?aid=1112642), submitted by fischi_bln

Program seems to work fine. All windows apear to be OK.
Just can't see any entries  in main window.

Main window continuously states: "wxMaxima 0.5.4
http://wxmaxima.sourceforge.net". After a "Clear
screen", main window still just shows "(%ixx) but
nothing more.

Input line works fine. Engine seems to work fine as
well, because it states :"Ready for user input",
"Maxima is calulating", "Parsing output" (etc.) when
calculating. 

Just can't see any entries in main window nor any
graphs at all. Following (shell) errors repeat again
and again:

---

*\* (wxmaxima:12841): CRITICAL **: file pango-layout.c:
line 2010 (pango_layout_get_extents): assertion `layout
!= NULL' failed

(wxmaxima:12841): GLib-GObject-CRITICAL **: file
gobject.c: line 1337 (g_object_unref): assertion
`G_IS_OBJECT (object)' failed

---

My  system: i686, P4, 2.4GHz, 512MB RAM
Linux: SuSE 9.1 (Standard), Kernel 2.6.5-7.111.19-smp
wxmaxima: version 0.5.4-1 (compiled by myself)
maxima: version 5.9.1-1 (rpm from
http://maxima.sourceforge.net)
wxGTK: version 2.4.2-138 (comes with SuSE 9.1)

Thanks for any hints!
Best regards,
Uwe alias Fischi_bln

PR #2318, opened 2026-09-18
Stop asserting on captured output from a GUI-subsystem Windows process

gunterkoenigsmann:
## Summary

The minGW job has been red on essentially every push since 2026-08-15, on `wxmaxima_version_string` alone. This gets it green, fixes two real bugs found on the way, and — the part worth arguing about — **narrows a test's platform coverage**. That last part is in tension with the "never weaken a test" rule added in #2317, so the case is made in full below rather than buried.

## The narrowing, and why I think it holds

The failing assertion is that text written by a **GUI-subsystem** process reaches a parent capturing it. That is a property of Windows, not of wxMaxima. An executable's subsystem is one bit in the PE header, fixed at link time; `wxmaxima.exe` must be GUI-subsystem or a console window appears beside every worksheet. That single bit is why `RedirectStdioToParent()` exists at all and why `wxmaxima-cli.exe` had to be a separate binary.

The investigation has disproven, with real evidence each time rather than by reasoning: the `FILE` struct copy, `_dup2()`'s handle ownership, ctest's `-j 2` scheduling, a ConPTY drain race, and a `cmd.exe` quoting bug in the workaround itself. The shipped traces show the process run, bind its streams, reach the `--version` branch, complete the write and exit 0 — while ctest captures nothing.

What tips it, for me, is #2317's own reasoning rather than an exception to it: a job red for a month is not a warning anyone still reads, and this one has been hiding whatever else the minGW runner had to say since 2026-08-15. **You may weigh that differently, and it's your call — the alternative is leaving it red until the root cause is found, which nobody has managed in a month.**

What is covered where now:

| | content asserted | exit code asserted |
|---|---|---|
| non-Windows | yes, and **stronger than before** | yes |
| Windows | **no** — stated at the point where the assertion would go back | yes, plus new `wxmaxima_cli_version_returncode` |

The Windows exit-code tests are worth more than they sound: before `wxMessageOutput` was pointed at stdout, a GUI-subsystem build answered `--version`/`--help` with a *modal message box*, so they'd hang to timeout rather than fail. That regression is still guarded.

## The assertion being dropped was weaker than it looked

`PASS_REGULAR_EXPRESSION "wxMaxima ${VERSION}.*"` referenced `${VERSION}` — a variable this project **never sets anywhere** (verified: no `set(VERSION ...)` in any `CMakeLists.txt` or `.cmake`; the version lives in `${WXMAXIMA_VERSION}`/`${PROJECT_VERSION}`). It expanded to `wxMaxima .*`, which is visible in the CI logs themselves as `Regex=[wxMaxima .*]`. So it never checked the version it is named after. Now fixed to `${WXMAXIMA_VERSION}` — confirmed in the generated `CTestTestfile.cmake` as a real `PASS_REGULAR_EXPRESSION "wxMaxima 26.08.0-dev"`.

Net: non-Windows coverage goes **up**.

## Two real bugs, fixed here

**1. `BindStdStreamToParent()` closed the standard handle it had just bound.** `_open_osfhandle()` makes the descriptor the *owner* of the handle, so the following `_close()` closed the handle `GetStdHandle()` still hands out — leaving a dangling `STD_*_HANDLE` whose value Windows then recycles. Whenever a parent points stdout and stderr at the same place (**any `2>&1`** — including `--logtostderr --batch ... 2>&1 | more`, and the redirect this very test used to use), binding stdout destroyed stderr's handle and all of stderr was silently lost.

This is what the `type=unknown` lines in the CI traces are; reading them as "ctest handed us an invalid handle" was wrong. Fixed by binding a `DuplicateHandle()` copy. Verified functionally under Wine with a console harness that creates a pipe the way kwsys does: **22 bytes captured before, 49 after**, deterministically.

**2. `wxmaxima-cli.cpp` passed its standard handles to `CreateProcessW()` as-is** rather than as explicitly inheritable duplicates. A handle named in `STARTUPINFO` only reaches the child's table if it carries `HANDLE_FLAG_INHERIT`, and nothing guarantees the ones our parent gave us do. Get it wrong and `CreateProcess()` still succeeds while the child finds handle *values* naming nothing it owns — **a mechanism that would explain the traces' `type=pipe`-in-parent / `type=char`-in-child finding.** Wine does not reproduce it, so this is a documented-contract fix, not a demonstrated cure; the launcher now logs `inherit=` so the next run's trace settles it.

## Also corrected

`AGENTS.md` twice named `wxMessageOutputStderr::Printf()` as the write under investigation, and proposed instrumenting the `fputs()` inside it. The `-v` branch is a plain `printf()`; `wxMessageOutput` is only used for `--help`'s usage text. That correction is in the entry, along with what to try first.

## Verification

- [x] Configure + build clean on Linux.
- [x] `wxmaxima_version_string`, `wxmaxima_version_returncode`, `wxmaxima_help_returncode` all pass locally (`3/3`), with the content check now asserting the real version.
- [x] `${VERSION}` confirmed unset repo-wide; generated CTest file confirmed to carry the literal version.
- [x] Bug 1 reproduced and its fix demonstrated under Wine, by byte count, not by reading code.
- [ ] Bug 2 is unverified as a *cure* — Wine can't reproduce the failure it targets. Stated as such rather than claimed.

Root cause of the capture failure itself remains unknown. Nothing here claims otherwise, and `AGENTS.md`'s entry is headed accordingly rather than marked resolved.

---


gunterkoenigsmann:
**`Compile using minGW` is green — `100% tests passed out of 134`.** First green minGW since 2026-08-15. Every other check on this PR is green too.

The run's own trace then answered the two questions this PR left open, and **both answers contradict claims I made in the PR description above.** Pushed as `cd99e10`, correcting the `AGENTS.md` text rather than leaving it to mislead whoever reads it next.

**1. The launcher's handles were already inheritable — so fix #2 is not the pipe→char mechanism.**

```
[wxmaxima-cli pid 7564] start: stdin handle=...318 type=pipe inherit=yes,
                               stdout handle=...3a8 type=pipe inherit=yes,
                               stderr handle=...30c type=pipe inherit=yes
[wxmaxima-cli pid 7564] CreateProcessW ok, child pid 3992
```

CTest sets `HANDLE_FLAG_INHERIT` itself, so there was never a cleared flag to lose — and child 3992 *still* reports all three as `type=char`. Passing explicitly inheritable duplicates changed the child's reported handle types **not at all**. I described this in the PR body as "a mechanism that would explain the traces' finding" and "the most promising remaining lead"; it is neither. The fix stays — MSDN's redirected-child example documents it, and relying on a flag nobody guarantees is a latent bug regardless — but it is not a cure.

**2. The `type=unknown` lines are unmoved by fix #1 — so that fix does not explain them either.**

Same 264 `BindStdStreamToParent(...) start:` lines in both runs:

| | `type=unknown` |
|---|---|
| before the fix ([#2315's run](https://github.com/wxMaxima-developers/wxmaxima/actions/runs/35337996797/job/105577197996)) | **37** / 264 |
| after the fix (this PR's run) | **36** / 264 |

Had a previous stream's binding been closing the handle, this would have gone to zero. It didn't move. So the reading the commit talked itself out of — that the child is handed an already-unusable handle — was very likely right all along, and "the previous binding closed it" was wrong. Post-fix the unknowns sit almost entirely on `STD_ERROR_HANDLE` (32 of 36), noted as a starting point.

**The bug that fix addresses is still real** — demonstrated under Wine by byte count, 22 → 49 — and still worth fixing. What fails is only its claimed *connection to those trace lines*. Those are two different claims and I ran them together: evidence that code is wrong is not evidence that it caused a particular symptom. Checking cost one grep of a log that already existed.

Net effect on what this PR is: the CI-hygiene case and the non-Windows coverage improvement are unaffected and stand as described. The "two real bugs" are still two real bugs. What's gone is the suggestion that either one explains the underlying capture failure — **the root cause is still unknown, and both of the leads this PR proposed are now closed.** The entry says so, and says to start from the unexplained fact itself.

---

PR #2324, opened 2026-09-21
Ask an AI provider which models it offers, and make Model a combobox

gunterkoenigsmann:
You were right that there is a pseudostandard here, and it turns out to be the thing that answers the objection this repo already had on record against doing this.

Two earlier follow-ups in `AGENTS.md` argued against model auto-detection, on the grounds that "model auto detection feels like hunting a mechanism that catches outdated strings with a mechanism that can get outdated" — a mechanism that would need maintaining against four APIs just to answer a question a link to each provider's docs already answers. That objection is *answered* here rather than overruled: there is no per-provider list-fetching code to maintain, because `GET <base>/v1/models` is the same de facto standard as the chat endpoint these providers already implement. A provider this codebase has never heard of — OpenRouter, Groq, Together, vLLM, LM Studio, llama.cpp server, a local Ollama — is covered by the same default that covers OpenAI, because it is the same URL rule.

## What changed

- **`AiProvider::ModelsRequestUrl()`** derives the list endpoint from the chat endpoint the user has already configured (drop a trailing `/chat/completions`, append `/models`) rather than storing a second URL that could get out of step with the first. Anthropic overrides it for `/messages` → `/models`; Google for its own `/v1beta/models`.
- **`AiProvider::ParseModelList()`** is a pure function with no I/O, unit-tested directly — the same split `BuildRequestBody()`/`ParseReply()` already use. Anthropic's list turns out to use OpenAI's exact `{"data":[{"id":…}]}` shape, so only Google needs its own: it reports `{"models":[{"name":"models/…"}]}`, whose `models/` prefix has to come off (the chat URL is built as `<base>/<id>:generateContent` and `<base>` already ends in `models/`, so leaving it on produces `models/models/`), and it lists models this provider cannot use, so entries that do not advertise `generateContent` are dropped. An entry that says nothing either way is kept.
- **`AiProvider::FetchModels()`** mirrors `SendChat()` deliberately — the unique `wxWindowIDRef`, the `shared_ptr` capture, the `State_Unauthorized` case. It has its own guard against the *second* terminal event that cancelling a 401 raises (`State_Cancelled` for the same id), which would otherwise fire the caller's callback twice for one fetch, the second overwriting the real reason with "cancelled".
- **The Model control is a `wxComboBox`, not a `wxChoice`** — the list is an offer, not a constraint. A model that is not in the list can still be typed; a fetch only ever replaces the dropdown's *items* and restores whatever is in the text part; a provider whose list is unreachable behaves exactly as the plain text field it replaced.
- **Nothing fetches automatically. A "Fetch models" button beside the combobox is the only trigger.** Most of these list endpoints want an API key, so fetching on provider selection would hand a third party the user's key — possibly one pasted seconds earlier — as a side effect of picking an entry from a dropdown, which nobody means as "contact them now". Because every fetch is therefore a deliberate press, every outcome is reported rather than swallowed; a press that showed nothing would be a bug in its own right. The button's tooltip says plainly that pressing it goes to the network using the key above.
- `wxTheApp` is the event sink and a `shared_ptr<bool>` says whether the dialog is still alive, because `wxWebRequest` holds a raw `wxEvtHandler*` and cannot learn that the dialog it was handed has been closed.

## What was verified, and what was not

**Verified live:** the Anthropic and Google model endpoints were asked directly and answered with their own "needs a key" errors — `https://api.anthropic.com/v1/models` → 401 `x-api-key header is required`, Google's `v1beta/models` → 403 `Please use API Key`. That proves both routes exist and are spelled as documented.

**Not reachable from this sandbox**, and so resting on the documented standard plus unit tests: OpenAI, Groq and a local Ollama (egress-blocked).

**Not verified at all:** the UI itself. This sandbox's wxWidgets has `wxUSE_SECRETSTORE` off, so the AI Chat tab does not exist in this build and I could not screenshot the combobox, the button or the status line. That's the same gap every earlier AI-chat change in this repo has carried; the widget construction is verified by reading it against wxWidgets' own contracts — including keeping the detail grid at 4×2 by putting the button in a `wxBoxSizer` beside the combobox rather than in a fifth row, since that sizer fixes both counts and asserts on the overflow. Worth a real look on a build that has the tab.

`test_AiProvider` covers URL derivation for every built-in provider, a local Ollama server and a non-standard base; both response shapes; Google's prefix stripping and `generateContent` filtering; and malformed bodies (a broken body throws, one bad entry does not discard the rest of the list). 194 assertions in 14 cases. Confirmed non-vacuous by mutation — removing Google's prefix strip fails two assertions.

Full suite passes. Builds clean under `-Wall -Wextra -Wpedantic`, and with both `-DWXM_USE_AI_TOOLS=ON` and `=OFF`.

## Known gap, not addressed here

A local server put behind HTTP Basic auth (which Ollama's own docs suggest) is **not** supported by this or by the existing chat path: `OpenAiCompatibleProvider::AuthHeaders()` only ever sends `Authorization: Bearer <key>`, so a Basic-auth reverse proxy answers 401 and wxMaxima reports it as a rejected API key. Fixing it means an optional username on a custom provider and a `Basic base64(user:password)` header; since both `SendChat()` and `FetchModels()` go through `AuthHeaders()`, one change would cover both. That belongs in its own PR.

## One separate finding, for its own PR

While checking which providers answer a model-list request I found that **GitHub Models was retired on 2026-07-30** (closed to new customers 2026-06-16), per GitHub's own changelog. `AiProviderKind::GitHubModels` was added here on 2026-09-11 — six weeks after the service stopped existing, so selecting it cannot do anything but fail. Deliberately left out of this PR's scope; it wants removing or repointing at a migration target.


gunterkoenigsmann:
`compile_latest_and_test` went red on `75c3c23`, and the failure is not this PR's:

```
groupcell_test_stubs.cpp:24: error: ambiguating new declaration of 'MyApp& wxGetApp()'
test_WorksheetLayout.cpp:456: note: old declaration 'TestApp& wxGetApp()'
```

That job builds with `-DCMAKE_UNITY_BUILD=ON`, and the stub file plus the test files that link it each declare `wxGetApp()` with a different return type — fine while they are compiled apart, an outright error once a unity batch merges them into one translation unit. This PR touches none of those files; its only test change is 112 added lines in `test_AiProvider.cpp`, which built fine.

I confirmed it is pre-existing rather than assuming so: an unmodified `main` in a separate worktree, configured the way that job configures, fails identically.

Fixed properly in #2325 (`SKIP_UNITY_BUILD_INCLUSION` on the stub, so it stays its own translation unit — renaming either function would break the symbol substitution the stub exists to perform). I have ported just that one-line CMake change here as `12cf926` so this branch can go green without waiting on #2325 to merge; it becomes a no-op once `main` carries it. The documentation half of #2325 is deliberately not ported, so the two branches can't duplicate the same paragraphs.

Worth flagging plainly: this is fallout from my own earlier #2321, which switched CI to unity builds. The collision was already there and latent; that PR is what made it visible — so `main` has been red since, and #2325 is worth landing ahead of this one.

---

PR #2328, opened 2026-09-21
Support HTTP Basic authentication for custom AI providers

gunterkoenigsmann:
Ollama's documentation recommends putting an instance reachable from beyond localhost behind an authenticating reverse proxy, and the proxies it steers people to (nginx, caddy) authenticate with a username and a password. wxMaxima could not reach such a server at all: every provider shape authenticated through its own single header (`x-api-key` / `Authorization: Bearer` / `x-goog-api-key`) built from one stored secret, and nothing here had a notion of a username. The proxy answered 401 — which, per the `State_Unauthorized` handling documented in `AGENTS.md`, does not even arrive as a normal completion.

## What changed

A custom provider gains a **Username** field. When it is non-empty the provider sends `Authorization: Basic base64(user:secret)` instead of its own credential header, and the key field below relabels itself to **Password:** so it is visible what the one stored secret currently means. Anthropic-shaped providers keep sending `anthropic-version`, which is a protocol version marker rather than a credential.

The username row appears only for a Custom provider, next to the request-URL and API-style controls that are already Custom-only — a built-in provider's auth shape is implied by its kind and is not editable.

## The one-credential-slot decision

There is deliberately **one** credential slot per provider, not two. A non-empty username reinterprets the existing `wxSecretStore` entry as the Basic password rather than adding a second secret.

- **What this covers** is the case actually reported: a local model server with no upstream API key of its own, sitting behind a proxy that wants a password.
- **What it deliberately does not cover** is a keyed cloud provider behind a Basic-auth proxy, which needs two independent secrets at once (the proxy's password *and* the provider's own key) and therefore a second secret-store slot plus a second UI field. That is a real feature with real UI cost, and nobody has asked for it.

## Persistence

The username is persisted alongside the rest of a custom provider's configuration (the same JSON array as `name`/`shape`/`baseUrl`/`model`). The password continues to live only in `wxSecretStore`, never in plain config. An entry written before this existed parses with an empty username and behaves exactly as it did before.

## Verification

- `test_AiProvider`: **198 assertions in 14 test cases**, all passing. Two new scenarios cover Basic auth across all three provider shapes (asserting the exact base64 literals, cross-checked independently rather than read back out of the implementation), that an empty username leaves every existing header byte-for-byte unchanged, that a colon inside the password is transmitted correctly (RFC 7617 splits on the first colon only), and a username config round trip including an entry with no `username` key at all.
- Full unit suite: **48/48** under `xvfb-run`.
- Clean build under `-Wall -Wextra -Wpedantic` with `WXM_USE_AI_TOOLS` both ON and OFF.

Not verified live: this sandbox's wxWidgets has `wxUSE_SECRETSTORE` off, so the AI Chat tab never opens here — the same caveat every other entry in this feature's history carries.

Two things worth flagging for review, both recorded in `AGENTS.md`:

- The Options grid's row count had to go from `wxFlexGridSizer(4, 2, …)` to `(5, 2, …)`. That is not cosmetic — a grid sizer with both dimensions fixed hard-caps its item count at `rows*cols` and asserts past it, which is a bug this feature has already hit twice and which this sandbox cannot catch.
- Adding a field to `AiCustomProviderConfig` breaks existing aggregate initializers under `-Wmissing-field-initializers`, which the `-Werror` jobs turn into a hard failure while a plain local build stays silent. Three initializers needed fixing; they were caught locally only because that warning was enabled deliberately.

---


gunterkoenigsmann:
Two separate CI failures here, both now addressed.

**1. The unity-build stub collision** (`compile_latest_and_test`, `compile_and_test_sanitized`, `compile_latest` on `72c8e26`) — `ambiguating new declaration of 'MyApp& wxGetApp()'` inside `Unity/unity_0_cxx.cxx`. That was fixed on `main` by #2325's `SKIP_UNITY_BUILD_INCLUSION`, which landed after this branch was cut. Merging `main` in (`b360849`) brought it in. Verified locally rather than assumed: configured the merged branch with `-DCMAKE_UNITY_BUILD=ON -DCMAKE_UNITY_BUILD_BATCH_SIZE=16` and built the exact targets CI failed on — clean, and `test_EditorCellBidi` and `test_AiProvider` both pass from that unity build.

**2. `check-pot-up-to-date`** — this one is partly mine and partly `main`'s, and this PR genuinely needed a fix of its own.

`main` is missing ten strings the model-list work (#2324) left out of `wxMaxima.pot`; #2330 fixes that. But **this branch adds three more on top**:

```
> msgid "Username:"
> msgid "Password:"
> msgid "Leave empty unless this endpoint sits behind an HTTP Basic authentication proxy. …"
```

So it would have failed the check on its own merits even after `main` is fixed. `31091f0` regenerates the POT here, which covers all thirteen. The overlap with #2330 is harmless by construction — both are `update-locale` output against the same sources, so whichever lands first leaves the other with nothing to add.

Only the POT is committed, per `AGENTS.md`: `update-locale` also rewrites fifty `.po` files with shifted line numbers, which would bury this under a six-figure diff. I checked rather than assumed that deferring costs nothing — re-running `msgmerge` later wipes **zero** translations and fuzzies **zero**, measured msgid-keyed against the current `de.po`.

Verified by reproducing CI's exact sequence locally (configure → `update-locale` → its own `strip_pot` diff against `HEAD`): fails before the commit, `PASS` after. Full unit suite 48/48, clean build.

---

PR #2329, opened 2026-09-21
Let an AI see and change which sidebars are visible, via MCP

gunterkoenigsmann:
Picks up the backlog item you scoped as "for the next branch and PR": let an AI query which sidebars are currently visible, and show/hide them.

Three tools:

| tool | what it does |
|---|---|
| `list_sidebars` | every dockable pane: stable `name`, human-readable `caption`, `visible`, `can_hide` |
| `show_sidebar` | show one, by a `name` from that list |
| `hide_sidebar` | hide one, same |

So an AI can open the Variables pane before discussing variables, or clear a sidebar out of the way, instead of asking you to do it.

## Why these count as safe writes

The backlog entry flagged one thing to settle first: whether toggling a sidebar belongs with `watch_variable`/`unwatch_variable`, or in the category the "no write/evaluate tools" rule exists to keep out. **I've put it in the former**, and the reasoning is the one that section already uses — what is the blast radius if the model gets it wrong? Here it is "a pane appeared or disappeared": visible to you the instant it happens, undone with one click, and unable to reach worksheet content, insert or edit a cell, or make Maxima evaluate anything. That is strictly *less* reach than `watch_variable`, which at least causes Maxima to be sent a query.

**This is a judgement about this kind of write, not a general loosening** — `insert_cell`/`evaluate_cell` remain the separate, much bigger decision. Say the word if you'd rather these were read-only and I'll drop the two write tools and keep `list_sidebars`.

## Plumbing

Which panes exist and which are shown lives on `wxMaximaFrame`; `McpTools` only ever holds a `Worksheet` and a `Variablespane`. So it's a callback pair, the same shape as the existing Maxima connection check.

Both halves go in **one** `SetSidebarAccess()` call on purpose: a build wired for enumeration but not toggling would advertise half-working tools, and nothing at the tool layer could tell that apart from a real failure. Left unset — the default — all three tools throw saying so, rather than returning an empty list that reads like "this wxMaxima has no sidebars."

Wired in `wxMaximaFrame`'s own constructor, not `wxMaxima`'s where `SetConnectionCheck()` goes, because `m_sidebarNames`/`m_sidebarCaption` are private to the frame and a derived class can't read them. Both lambdas re-read the map per call, so panes registered later in that same constructor are still reported correctly.

## Two things the frame's API swallows, caught before reaching it

- `ShowPane()` **silently does nothing** for a name it doesn't know — so handing it an unchecked name would report success for a typo. The tools look the name up first.
- `ShowPane()` **asserts** if asked to hide the worksheet (wxAUI's centre pane). `hide_sidebar` refuses `console` before that point, and `list_sidebars` reports `can_hide: false` so it never has to be found by failing.

Only the untranslated name is accepted as an identifier. The caption is reported too, but an AI keying off it would break the moment wxMaxima runs in another language — so the unknown-name error lists the real names and says so explicitly.

## Verification

**Unit** — `test_McpTools`: 146 assertions in 10 test cases. Two new scenarios cover listing, show, hide, show-when-already-shown, the `console` refusal, an unknown name, a caption used as a name, a missing argument, the unset-access case, and `CallTool()` dispatch. They run against a stand-in for the frame (a real `wxMaximaFrame` is far too heavy for a unit test, and the callback design is what makes one unnecessary).

**Live** — since that stand-in is by definition not the real frame, the frame half was checked against a running app in Xvfb with the MCP server enabled:

- `list_sidebars` returned all **17** real panes, sorted, with `console` correctly `can_hide: false`
- `show_sidebar variables` and `hide_sidebar structure` both took effect — confirmed by reading `list_sidebars` back **and** by a screenshot: Variables docked at the bottom, Table of Contents gone, every other pane untouched
- hiding `console`, an unknown name, and a caption-as-name each returned a JSON-RPC error with the intended message

Full unit suite 48/48; clean build with `WXM_USE_AI_TOOLS` both ON and OFF (MCP isn't gated by it, but I touched `wxMaximaFrame.cpp` near that guard).

One test bug worth flagging since it cost real time: `for (const auto &t : tools.ListTools()["tools"])` iterates a json that has **already been destroyed** — the temporary isn't lifetime-extended, because `operator[]` yields a reference *into* it rather than the temporary itself. It failed as though the tools were never registered, while the source plainly showed they were, which sends you looking at the build rather than the test. Fixed, documented, and a sweep found no other instance of that shape in the tree.

Docs: the backlog item is removed from `AGENTS.md` and the reasoning above recorded in the `wxmaxima-ai-tools` skill, which is where the MCP notes moved.

---


gunterkoenigsmann:
`compile_latest_and_test` is red here, but not because of this PR.

The failing step is `check-pot-up-to-date`, and it is failing on `main` itself — the model-list fetch feature (#2324) added user-facing strings without re-running `update-locale`, so `wxMaxima.pot` is missing ten of them. Every branch cut from `main` inherits it:

```
$ git show origin/main:src/dialogs/ConfigDialogue.cpp | grep -c "Fetch models"
1
$ git show origin/main:locales/wxMaxima/wxMaxima.pot | grep -c "Fetch models"
0
```

This PR adds **no** translatable strings at all — the MCP tool descriptions and error messages are protocol text sent to an AI client, never shown in the UI, so none are wrapped in `_()`. There is therefore nothing to fix on this branch: it has no POT drift of its own, and the check will pass here once `main` is fixed.

I've opened #2330 for that fix rather than carrying it in this PR — per `AGENTS.md`, an unrelated fix found while a PR is open is its own branch and its own PR.

I have not re-run the check, since a re-run cannot pass while the cause is still on `main`; #2330 merging is what clears it.

---

